Issue #263, opened 2015-01-07
PropAllDiffInst filtering could be improved


cprudhom:
See test AlldifferentTest#testB()

cprudhom:
I suggest to split the propagator in two: one (the current one) dedicated to enumerated domains, and a second one dedicated to bounded domains (i.e. when at least one domain is bounded).
The second one reacts on bound events and stores, in a backtrackable list, the indices of already instantiated variables to better the filtering.

cprudhom:
Tests shows that separating into two propagators is a good choice: it reduces the number of nodes explored when dealing with bounded domains.
However, when the number of nodes is not reduced (when domains are totally disjoint for instance), it is a slower than the previous one, due to a double loop.
I'll post a PR.

cprudhom:
See PR #286 for propal

cprudhom:
The expected behavior is to only react on instantiation. 
If we want to have a better algorithm for bounded variables, simply use the BC one.

PR #286, opened 2015-04-28
Fix issues #263

cprudhom:
Fix issues #263


choco-solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/PropAllDiffInst2.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+
+    //***********************************************************************************
+    // INFO
+    //***********************************************************************************
+
+    @Override
+    public int getPropagationConditions(int vIdx) {
+        return IntEventType.boundAndInst();
```
cprudhom: If there is at least one enum variable...

cprudhom:
What about running it with BC?

cprudhom:
It is redundant with BC

PR #300, opened 2015-05-18
Request for comments: Remove monitors from the solver after optimal solution is found.

JLiangWaterloo:
Many meta-algorithms require calling the solver many times, for example, the algorithm in the paper "Time-versus-capacity compromises in project scheduling" requires calling many (> 100,000 in some cases) single-objective optimization searches. In the current implementation, every single-objective optimization attaches two permanent monitors. If we need to call single-objective optimization many times, then the monitors accumulate and the performance degrades.

Perhaps even worse, one of the monitors attached will always restore solution on close which makes it impossible to do a satisfaction search after a single-objective optimization. For example, in the following code, the satisfaction part will only find one solution (i.e., misses 10 solutions). If code was rewritten to do the satisfaction part first, then it will not miss any solutions.

```java
        Solver s = new Solver();
        IntVar i = VF.enumerated("i", -5, 5, s);

        s.findOptimalSolution(ResolutionPolicy.MAXIMIZE, i);
        System.out.println("Optimization problem: " + i);

        s.getEngine().flush();
        s.getSearchLoop().reset();

        if (s.findSolution()) {
            do {
                System.out.println("Satisfaction problem: " + i);
            } while (s.nextSolution());
        }
    }
```


choco-solver/src/main/java/org/chocosolver/solver/Solver.java
```diff
@@ -873,8 +889,15 @@ public void findOptimalSolution(ResolutionPolicy policy, RealVar objective, doub
         if (!getObjectiveManager().isOptimization()) {
             set(new ObjectiveManager<RealVar, Double>(objective, policy, precision, true));
         }
-        set(new LastSolutionRecorder(new Solution(), true, this));
-        solve(false);
+        ISolutionRecorder oldRecorder = getSolutionRecorder();
```
cprudhom: What is the point in storing the previous solution recorder ?

JLiangWaterloo: I didn't want to leave a stopped LastSolutionRecorder attached to the solver so I gave it back it's old one. Perhaps I should remove lines 894 and 898 and not touch the solvers solutionRecorder field.

cprudhom: Well, if one manually declared a solution recorder, it does not seem a good idea to me to unplug it just before running the search.
Is the idea behind the patch to avoid using the wrong solution recorder when switching from optimization to satisfaction?

cprudhom:
Maybe a simple way to process is to remove the solution recorder in the `s.getSearchLoop().reset();` ?

jgFages:
sounds like a good idea

JLiangWaterloo:
There are search monitors created elsewhere, for example this [line](https://github.com/JLiangWaterloo/choco3/blob/unplug-finished-monitors/choco-solver/src/main/java/org/chocosolver/solver/Solver.java#L182) will plug a monitor on the Solver's constructor. Should `s.getSearchLoop().reset()` remove this monitor too?

choco-solver/src/main/java/org/chocosolver/solver/Solver.java
JLiangWaterloo: The idea behind the patch is to clean up the search monitors the findOptimal algorithm uses especially since the monitors it plugs alters the behaviour of the solver (i.e., it's monitor will "restore on close").

I modified my pull request to not touch the pre-existing search recorder.

jgFages:
It would be nice to have an overview of monitors we want to keep and monitors we want to remove after reset...

cprudhom:
Apart from the types of `ISearchMonitor`, the goals of the ones available in choco-solver twofold:

1. observing the search:
  i. for strategies (ABS, IBS, dom/wdeg, LC, CBJ, DBT and LNS+neighborhoods)
  ii. for reporting (MeasureRecorder, Chatterbox, NogoodFromRestarts, ACounter and SolutionRecorder) 

2. interacting with the search:
  i. to limit the search (ACounter)
  ii. to overpass the default behavior of the search loop (RestartManager, based on ACounter too)

Some of them have specific status:
a. measure recorder: created once for all on `Solver` initialization, can't be removed without throwing an error (measure is used in very numerous places),
b. solution recorders: should be attached once for all at `Solver` initialization but the type of resolution called affects the type of solution recorders required (or wanted).
c. limits: when reached they have to be reset otherwise the search can't go on.

All non-user defined ones can't be removed easily. For instance, when a search strategy, ABS, is overridden, its search monitors are kept.  The same goes for solution recorders.

Maybe, a first step would be to consider `SolutionRecorder`s as specific SM, such as `MeasureRecorder`. For instance, we could impose that only one `SolutionRecorder` can be attached to the search loop at a time?

cprudhom:
BTW, I partially add your patch in my current branch (all about unplugging monitors).
It will be available on the next merge, so we can still think about a good way to deal with monitors.

cprudhom:
As I revamp the search loop, I also reduce the role of ISearchMonitor to observe the search loop, they are not allowed (theoretically) to interact with the search anymore.
I changed the way search monitors are handled:
1. they are added to a list within a Solver on a call to `Solver.plugMonitor(ISearchMonitor sm)`
2. on a call to `Solver.solve(boolean)` (called anytime a resolution is needed):
a. the search loop search monitors are all removed,
b. all search monitors of the solver are transferred into the search loop.

A call to `Solver.unplugMonitor(ISearchMonitor sm)` will remove a single search monitor from the list of SM to be transferred to the search loop on the next call to `solve(boolean)`.
A call to `Solver.unplugAllMonitors()` will clear the list of SM to be transferred to the search loop.

Even if it is still not perfect, (for instance, when activity-based search is declared, the search monitor it relies on may be detached by error), I believe that it almost a more flexible way of dealing with search monitors.

Note that the same goes with the list of `Criterion` (which stop the search loop).

JLiangWaterloo:
The specific problem I had in the original post still exists. That is:

```java
    Solver s = new Solver();
    IntVar i = VF.enumerated("i", -5, 5, s);

    s.findOptimalSolution(ResolutionPolicy.MAXIMIZE, i);
    System.out.println("Optimization problem: " + i);

    s.getEngine().flush();
    s.getSearchLoop().reset();

    if (s.findSolution()) {
        do {
            System.out.println("Satisfaction problem: " + i);
        } while (s.nextSolution());
    }
}
```

Will print
```
Optimization problem: i = 5
Satisfaction problem: i = 5
```

Even though I expect there to be 11 solutions to the satisfaction problem instead of just the one.

cprudhom:
I changed a little bit your test to:
```java
Solver s = new Solver();
        IntVar i = VF.enumerated("i", -5, 5, s);
        s.findOptimalSolution(ResolutionPolicy.MAXIMIZE, i);
        Assert.assertEquals(s.getMeasures().getSolutionCount(), 1);
        Assert.assertEquals(s.getSolutionRecorder().getLastSolution().getIntVal(i).intValue(), 5);

        s.getEngine().flush();
        s.getSearchLoop().reset();

        s.findAllSolutions();
        Assert.assertEquals(s.getMeasures().getSolutionCount(), 11);
```

And it does not fail one the current develop branch.

By the way, in order to be less intrusive, I removed the code from all `ISolutionRecorder` which automatically restored the last (best) solution into the solver on ending the search.
Now an explicit call to one of the followings is required: 
```java
solver.restoreLastSolution();
// or
Solution aSolution= ...;
solver.restoreSolution(aSolution);
```

Issue #304, opened 2015-06-01
Solver interruption: add priorities


cprudhom:
In order to deal with fast restart in LNS and external clean interruption (using `solver.getSearchLoop().interrupt()`) there should be interruption priorities to stop correctly the resolution.

cprudhom:
There are (at least) 5 reasons to interrupt the search:
1. a *limit* is reached,
2. one solution is asked (by calling `findSolution()` or `nextSolution()`) and one *solution* is found,
3. the *root node* is reached
4. *inconsistency* is detected at root node during initial propagation or on restart posting the cut (in COP),
 5. an *external* request is done.

+ 1, 2, 3 and 4 are 'weak' reasons: the search can be resumed without specific condition. For instance, LNS can overpass the interruption instruction: 3 can be produced with small fragments and 4 with incorrect fragments, in that cases the search can be resumed
+ 5 is a 'strong' reason: the search can't be resumed with a call to `searchloop.reset()`.

Note that weak reasons can be encountered again and again if no modification is made to the model or the search strategy. For instance, 'normal' backtrack search interrupted because of 3 can be resumed but will then be interrupted because of 3 again if nothing has changed since the last call.

In consequence, I suggest to add a boolean parameter to the `ISearchLoop.interrupt(String message)` method stating whether or not the interruption is strong. By default, it is set to `false`(weak).

jgFages:
From what I understand, a strong interruption should kill the solving process. However, it should remain possible to resume it MANUALLY (even after a strong interruption). Do you agree?

example: Imagine I want to call findOptimalSolution with LNS for 1 min, then change a few things in the model and continue optimization. It should be possible to resume the search.

chocotrust:
Sure.  I considered this in my proposal.

cprudhom:
https://gist.github.com/506903e1653c0a764cff.git

cprudhom:
See PR #307

cprudhom:
So, any comments before I merge it?

cprudhom:
Well, finally, I think this is still not enough, in particular when dealing with `Portfolio.nextSolution()`: how to make sure a worker has not ended because it has explored the entire search space?

cprudhom:
I delt with the various status search can encounter: 
- does search end normally,
- is it complete,
- does it have reach a limit,
- is it resumable.

There is still an issue with Portfolio, when a worker has a limit (but, for the moment, workers are not accessible easily...).

cprudhom:
[...]

PR #307, opened 2015-06-04
Issues #304:

cprudhom:
- modify ISearchLoop.interrupt(), add new parameter 'avoidable' to qualify the strength of the interruption


jgFages:
Are lines 92 and 93 of LargeNeighborhoodSearch.java necessary?

jgFages:
sounds like "true" means "strong", which is in contradiction with the word "voidable", right?

chocotrust:
Not correct: set `voidable` to `false` (ie, strong) set `alive` to false and stop the process. Moreover, `canBeResumed()` will then return `false` which is the expected behavior.

PR #340, opened 2015-10-14
Revamping SearchLoop

cprudhom:
I'm considering revamping the `SearchLoop` object.
It seems clear that the current version has some strong limitations:
- there is no easy and correct way to interrupt the search when dealing with LNS and/or multi-thread resolution,
- there is an abusive use of `ISearchMonitor`s, which were initially created to observe the search loop, now most of them hack it (interruption interception for instance),

I already implement on my own branch an adapted version of ["Unifying search algorithms for CSP", Jussien and Lhomme](http://www.emn.fr/z-info/jussien/publications/jussien-RR0203.pdf) which is very flexible. I was able to re-implement existing stuff and implement ["Limited-discrepancy Search", Harvey and Ginsberg](http://www2.informatik.uni-freiburg.de/~ki/teaching/ss10/readinggroup/harvey-ginsberg-ijcai1995.pdf) and ["Depth-bounded Discrepancy Search", Walsh](http://ijcai.org/Past%20Proceedings/IJCAI-97-VOL2/PDF/086.pdf) really easily.
The code of LNS has been revisited too, 
I also changed the way limits are handled, using Java8 lambdas.

The branch is still under development and all suggestions/modifications are welcomed !


cprudhom:
Done directly in develop

PR #343, opened 2015-11-04
Fixed BitsetArrayIntVarImpl::updateUpperBound adding spurious values to its delta

JLiangWaterloo:
For example:

```java
public static void main(String[] args) throws Exception {
    Solver s = new Solver();
    IntVar i = VF.enumerated("i", new int[]{0,98,99}, s);
    IIntDeltaMonitor d= i.monitorDelta(Cause.Null);
    i.updateUpperBound(98, Cause.Null);
    d.freeze();
    d.forEachRemVal((IntProcedure) System.out::println);
    d.unfreeze();
}
```

This will print "99 98 0" even though the only removed value is 99.


cprudhom:
I added a test and also patched `updateBounds()`.

PR #378, opened 2016-03-17
Update LearnDBT to remove unneeded looping over the decision path

mwahbi:
The jumping target can be obtained directly from (last - nworld) instead of looping over the decision path.


cprudhom:
On my test suite, it throws your patch throws an error.

cprudhom:
Well, I introduced a bug somewhere else, so, forget my last comment.
Thank you.

cprudhom:
The PR was added in previous commit. Thank you for contributing.
Two questions though: 
- is your employer ok with you providing code to choco ?
- if ok, can I add your name to `choco-solver/CONTRIBUTORS` ?

mwahbi:
@cprudhom No problem, you can do it.

PR #393, opened 2016-04-25
when a thread used by choco solver is interrupted by Thread.interrupt() method, the solver should be stopped

gregy4:
Minimal change in develop branch to support Thread.interrupt() method used by thread pools to end a thread since method Thread.stop() is deprecated.

Maybe extra property of Solver class may be created to know whether a solver was stopped thanks to interrupted thread or some stop criteria was met.


cprudhom:
I'll merge your PR, two questions though: 
- is your employer ok with you providing code to choco ?
- if ok, can you add your name to `choco-solver/CONTRIBUTORS` ?

PR #396, opened 2016-04-28
Improve FirstFail

fhermeni:
I got some performance issues when having tons of variables.
This is a port of my own FirstFail that provided the following optimisations:

- Add a shortcut to stop at the moment we have a variable with a domain size of 2
- store the loop offset

Performance  improvement on my code:
```
Vanilla FirstFail: After 66897ms of search (terminated): 6439 opened search node(s), 1817 backtrack(s), 1 solution(s):
New FirstFail: After 55937ms of search (terminated): 6439 opened search node(s), 1817 backtrack(s), 1 solution(s):
```

### DIFF
--- a/src/main/java/org/chocosolver/solver/search/strategy/SearchStrategyFactory.java
+++ b/src/main/java/org/chocosolver/solver/search/strategy/SearchStrategyFactory.java
@@ -261,7 +261,7 @@ public class SearchStrategyFactory {
      * @return assignment strategy
      */
     public static IntStrategy minDomLBSearch(IntVar... vars) {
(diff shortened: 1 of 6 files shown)


cprudhom:
I would appreciate if we could avoid adding Model as input for the constructor.

cprudhom:
Ok, I couldn't find any easy way to avoid passing the `model` as parameter without breaking everything around.

fhermeni:
Can't you catch if from the first variable ?

cprudhom:
No, I can find wrong case: consider two strategies in sequence, the second may start at incorrect index when the first backtracks too high.

PR #400, opened 2016-05-02
Feature multicriterion

glelouet:
again sorry for format.

two functions to add :
Solution org.chocosolver.solver.search.IResolutionHelper.findOptimalSolution(IntVar[] objectives, boolean maximize, Criterion... stop)

no comment on this because I know you like to make your own comment. objectives can be null, stop can be null.

the test in 
org.chocosolver.solver.search.TestMultiSequentialObjectives.simpleSumTest()


cprudhom:
I would prefer a patch instead of this PR that seems very strange to me (did you encountered troubles while branching develop ?) 
Moreover, even if I like java-documenting methods myself, some clues are appreciate 👍

cprudhom:
Ok, I got it, you pull to `master`. So please, try again on develop, I delete this PR.

PR #401, opened 2016-05-02
Feature multicriterion

glelouet:
multi-objective while searching


cprudhom:
Thank you for providing this pull request.
I adapted your code to avoid posting/unposting the constraint on each solution.
In consequence, I will commit directly on `develop` and close this PR.

Issue #397, opened 2016-04-28
Tuning the TrailingEnvironment


fhermeni:
One the biggest problems I am dealing with, resizing the stack is the hot CPU path:
12,1% - 2 702 ms org.chocosolver.memory.trailing.trail.flatten.StoredIntTrail.resizeUpdateCapacity

It would be great to be able to customize the initial stack size and the growing factor to tune the solver when needed

fhermeni:
I planned to extend it for customization but it is final so it limits the possibility to play with it properly

cprudhom:
To allow anybody tuning it, I suggest to externalize its configuration.

fhermeni:
In details, it would be great to be able to indicate a specific trail size per variable or backtrackable stuff.

In my case, at least when I do `getEnvironment().makeInt()`, I would be able to set somewhere the initial size of the backed array.

fhermeni:
You can simplify EnvironmentTrailing now I guess ? It still proposes its default values while the builder takes care of that properly now.

cprudhom:
Sure. I'll do it

jgFages:
Hi Fabien, any feedback about the different environment performances on your benchmarks?

fhermeni:
it works well on my case.
I am waiting from some support from @cprudhom about a method I cannot implement

jgFages:
I have completed EnvironmentBuilder so that it is easier to test the CHUNK, FLAT and UNSAFE implementation: ::

Model model = new Model(EnvironmentBuilder.buildChunkEnvironment(),"");

Could you test the three alternatives?

fhermeni:
What I  did with Charles is a custom chunk environment. 
I cannot test your code in develop because I am currently running with the released version of Choco and the APIs are incompatible

jgFages:
ok

cprudhom:
See #397, #405 and #421

PR #404, opened 2016-05-06
Address partially #397

fhermeni:
It is now possible to set a default stack size per trail type.

It is possible to do something better but it would be too invasive (the coupling between the trailing environment and the various store* implementation) seems strong.


src/main/java/org/chocosolver/memory/trailing/EnvironmentTrailing.java
```diff
@@ -39,6 +39,10 @@
 import org.chocosolver.memory.trailing.trail.unsafe.UnsafeIntTrail;
 import org.chocosolver.memory.trailing.trail.unsafe.UnsafeLongTrail;
 
+import java.util.EnumMap;
+import java.util.EnumSet;
```
cprudhom: Codacy Issue found: [Avoid unused imports such as 'java.util.EnumSet'](https://www.codacy.com/app/cprudhom/choco-solver/file/2528135068/issues/source?bid=3218989&fileBranchId=3282683#l43)

fhermeni:
Messed up with the previous one :/

PR #405, opened 2016-05-06
Customizable stack size for the trailing environment #397

fhermeni:
It is now possible to set a default stack size per trail type.

It is possible to do something better but it would be too invasive (the coupling between the trailing environment and the various store* implementation) seems strong.


cprudhom:
Ok for the PR, but why is there a reference to commit da42b9c4222e99ecf2d5bbd48911dbcd8e247be6 in it ?

fhermeni:
Don't know why. I was not able to get rid of it. But if you look at the changeset, it only modifies the environment.

cprudhom:
I created an environment builder that allows configuration of each trail.
I also added CHUNCK back.
So, have a look at EnvironmentBuilder, that should fit your needs (btw, I revamped your PR)

fhermeni:
See this alternative implementation that hide the concrete type of each trail.
The code is smaller and there is no more interleaving between the environment that the trails:

https://github.com/fhermeni/choco3/commit/8855e7c8a3119ce1d2a6aba634ccc10d0fa4e8df

cprudhom:
Better, smaller: pushed 9a1b0ebcf6bbde8f52894d2a3da0e8297ab8be74

PR #407, opened 2016-05-09
Set var.card

glelouet:
- added SetVarImpl. getCard() and setCard(IntVar)
 - added corresponding simple test
 - specified setVar(String name, int... values) instead  of setVar(String name, int[] value) with null test on values

### DIFF
--- a/src/main/java/org/chocosolver/solver/variables/impl/SetVarImpl.java
+++ b/src/main/java/org/chocosolver/solver/variables/impl/SetVarImpl.java
@@ -152,7 +152,7 @@ public class SetVarImpl extends AbstractVariable implements SetVar {
 	public void setCard(IntVar card) {
 		if(cardinality==null){
 			cardinality=card;
-			new Constraint("SetCard", new PropCardinality(this, cardinality)).post();
+			new Constraint("SetCard", new PropCardinality(this, card)).post();
 		} else {
 			model.arithm(cardinality, "=", card).post();
 		}
--- a/src/main/java/org/chocosolver/solver/variables/impl/SetVarImpl.java
+++ b/src/main/java/org/chocosolver/solver/variables/impl/SetVarImpl.java
@@ -32,7 +32,10 @@ package org.chocosolver.solver.variables.impl;
 import gnu.trove.set.hash.TIntHashSet;
 import org.chocosolver.solver.ICause;
 import org.chocosolver.solver.Model;
+import org.chocosolver.solver.constraints.Constraint;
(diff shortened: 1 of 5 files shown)


src/main/java/org/chocosolver/solver/variables/impl/SetVarImpl.java
```diff
@@ -130,6 +134,29 @@ public ISet getLB() {
 	public ISet getUB() {
 		return ubReadOnly;
 	}
+	
+	@Override
+	public IntVar getCard() {
+		if(cardinality==null){
+			int ubc =  ub.getSize(), lbc = lb.getSize();
```
cprudhom: Codacy Issue found: [Use one line for each declaration, it enhances code readability.](https://www.codacy.com/app/cprudhom/choco-solver/file/2534693622/issues/source?bid=3218989&fileBranchId=3284892#l141)

jgFages:
I have two remarks about "specified setVar(String name, int... values) instead of setVar(String name, int[] value) with null test on values"

1) The "value" of a set variable is a set of integers so the method should be named setVar(String name, int... value) (no "s" at the end of value). 

2) I see it is a bit more convenient to avoid writing new int[]{} thanks to the ... but I am not sure it is good. I fear a user mimics integer variables passing 0, 9 and thinking it will represent [0,9] instead of {0,9}, or even worse, all its subsets... What do you think?

glelouet:
1) It depends whether you consider the parameter is a single it array or an array of it, I don't really care. I'm used to using plural form when there are several elements, whether through a collection, an iterable, a stream, an array, which do not have their own behavior.

2) if a user wants to make an intVar as a setVar, we can't help him. SetVar and IntVar are very dissociated and trying to use one like another would surely lead to errors anyhow.
However I like the idea of constructing a setVar over a range (like python), what do you think of adding setVarOver(name, int min, int max) ?

jgFages:
2) Are you suggesting : 
setVarOver(name, int min, int max) returns a constant [min, max]
and 
setVarOver(name, int minLB, int maxLB, int minUB, int maxUB) return a variable?

glelouet:
SetVar IVariableFactory.setVarOver(String name, int min, int max) returns a new SetVar with array of values fixed to {min..max}

jgFages:
good

(I did not want it to create a variable that is a subset of {min...max}. The variable should specify both the LB and the UB)

glelouet:
https://github.com/chocoteam/choco-solver/issues/411 ?

jgFages:
I wait #411 to be over before merging this PR

PR #421, opened 2016-05-25
More idiomatic pre-packaged trailing environment

fhermeni:
EnvironmentTrailing is still using a flat environment by default which is redundant with the purpose of the builder.


cprudhom:
I added it manually

PR #441, opened 2016-08-13
Add a method to check if a SetVar has a cardinality variable.

JLiangWaterloo:


jgFages:
Can you please update it with current master and replace 
if(cardinality==null) of setCard() by if(!hasCard())? It will provide a better coverage
thanks

JLiangWaterloo:
Done.

PR #442, opened 2016-08-13
Record all variables in a solution by default.

JLiangWaterloo:
I use a lot of channeling constraints and only use one of the variables in the search strategy, but I would still like to query the values of the other variable.

For example:

```java
        Model model = new Model();
        BoolVar[] bools = model.boolVarArray("bool", 2);
        SetVar set = model.setVar("set", new int[]{}, new int[]{0,1});
        model.setBoolsChanneling(bools, set).post();
        solver.setSearch(Search.minDomLBSearch(bools));
        
        Solution solution = solver.findSolution();
        System.out.println(Arrays.toString(solution.getSetVal(set)));
    }    
```

This will throw a null pointer exception because the value for "set" was never recorded.

### DIFF
--- a/src/main/java/org/chocosolver/solver/Solution.java
+++ b/src/main/java/org/chocosolver/solver/Solution.java
@@ -102,7 +102,7 @@ public class Solution implements ICause {
         empty = false;
         boolean warn = false;
         if (varsToStore.length == 0) {
-            varsToStore = model.getSolver().getSearch().getVariables();
+            varsToStore = model.getVars();
         }
         assert varsToStore.length > 0;
         if (intmap != null) {


jgFages:
We had long discussions about this. 
At first I feared storing all variables would degrade performance for both cpu and memory. 
But well, as long as it is still possible to specify which variable to store...
Besides, we had to make some ugly workarounds such as adding a dummy search heuristic to store some other variables. I now think it makes more sense to store all by default.

jgFages:
Can you make the PR up to date with master? I have a warning...

JLiangWaterloo:
Done.

PR #443, opened 2016-08-13
Check map if null.

JLiangWaterloo:
For example:

```java
        Model model = new Model();
        IntVar a = model.intVar("a", 1);

        Solution solution = model.getSolver().findSolution();
        System.out.println(solution.getIntVal(a));
    }
```
This will throw a null pointer exception.


jgFages:
Thanks

PR #446, opened 2016-08-14
Fix ScaleView::doRemoveValueFromVar to not remove values it does not contain

JLiangWaterloo:
For example, consider the following program:

```java
        Model m = new Model();
        IntVar y = m.intVar("y", -3, 1);
        IntVar z = m.intVar("z", new int[]{-4, 4});
        m.arithm(m.intScaleView(y, 3), "!=", z).post();
        System.out.println(m);
        Solver s = m.getSolver();

        while (s.solve()) {
            System.out.println("3 * " + y.getValue() + " != " + z);
        }
    }
```

This will fail to find the solution 3 * -1 != -4


jgFages:
Thanks, I included it directly and added a unit test from your example

Issue #448, opened 2016-08-29
Better filtering for set intersection


JLiangWaterloo:
For example, consider the following example:

```java
        Model m = new Model();
        SetVar s1=m.setVar("s1", new int[]{1,2,3}, new int[]{0,1,2,3,4,5,6});
        SetVar s2=m.setVar("s2", new int[]{}, new int[]{0,1,2,3,4,5,6});
        Constraint c = m.intersection(new SetVar[]{s1,s2}, m.setVar());
        c.post();
        
        m.getSolver().propagate();
        System.out.println(s1+ " intersect " + s2 +" is empty");
```

The values 1, 2, and 3 should be removed from s2, but the current propagator does nothing.

jgFages:
Yes there is no filtering from the result to the input sets regarding values in some input LB and not in the result UB. The reason is that it could become very time consuming for getting no inference when intersecting many sets (it is allowed to have same values on n-1 sets, where n is the number of sets to intersect). I believe it would make sense to add a complementary propagator to do it in the binary case. Can you do it?

PR #449, opened 2016-08-30
A complementary intersection propagator for filtering the sets.

JLiangWaterloo:
This PR is with regards to issue #448.

I know in your message you said to specialize it for the binary case, but I need it for the n-ary case.

I've ran about a million fuzz tests for the intersection constraint. Based on these tests, I think PropIntersection and PropIntersectionFilterSets are both idempotent. In addition, I think the combination of these two propagators are bound consistent. I couldn't find a counterexample in the million fuzz tests.

### DIFF
--- a/src/main/java/org/chocosolver/solver/constraints/set/PropIntersectionFilterSets.java
+++ b/src/main/java/org/chocosolver/solver/constraints/set/PropIntersectionFilterSets.java
@@ -92,12 +92,11 @@ public class PropIntersectionFilterSets extends Propagator<SetVar> {
             ISetIterator iter = vars[i].getLB().iterator();
             while (iter.hasNext()) {
                 int j = iter.nextInt();
-                if (!intersection.getUB().contains(j)) {
-                    if (count.adjustOrPutValue(j, 1, 1) >= k - 1) {
-                        // intersection does not contain j but k - 1 sets
-                        // contains j, hence the last set must not contain j.
-                        findSetThatDoesNotContainJInLB(j).remove(j, this);
-                    }
+                if (!intersection.getUB().contains(j) &&
+                    count.adjustOrPutValue(j, 1, 1) >= k - 1) {
+                    // intersection does not contain j but k - 1 sets contains
+                    // j, hence the last set must not contain j.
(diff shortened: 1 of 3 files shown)


cprudhom:
Thank you for the new propagator.
In order to merge the PR, we also appreciate if you : 
- could add it to the constraint declaration signature in `ISetConstraintFactory`,
- and could integrate tests to the test suite ?

src/main/java/org/chocosolver/solver/constraints/set/PropIntersectionFilterSets.java
```diff
@@ -0,0 +1,122 @@
...
+            sdm[idxVarInProp].forEach(j -> {
+                if (!intersection.getUB().contains(j)) {
+                    SetVar uniqueSet = findUniqueSetThatDoesNotContainJInLB(j);
+                    if (uniqueSet != null) {
+                        uniqueSet.remove(j, this);
+                    }
+                }
+            }, SetEventType.ADD_TO_KER);
+        } else {
+            sdm[idxVarInProp].forEach(j -> {
```
jgFages: Does this create a new lambda at each propagation? Doesn't it have an overhead compared to creating an object once and for all in the constructor?

jgFages:
Thank you also about your remark on the idempotency of PropIntersection. I have to admit that I do not remember what made us think it was not idempotent. Maybe a bug which has been fixed since then. I believe you can replace the redundant second propIntersection propagator with your new propagator. However, if you remove the second propIntersection propagator, I would feel more confortable if the tests you will add to the suite also cover the idempotency question in addition to the correctness check. Thanks again

JLiangWaterloo:
Done.

I'm not sure how you feel about black box fuzz tests in your testing suite.

JLiangWaterloo:
I just wanted to add one comment: the fuzz tests I typically run for my project are more "random" than the ones in this CL for two reasons:

1. I use a random SetValueSelector for my own fuzz tests, which the Choco library does not have.
2. I have a special implementation of SetStrategy that randomizes whether to enforce-first or remove-first per decision. The current implementation of SetStrategy cannot randomly switch between enforce-first and remove-first since it must stick with the choice passed through its constructor.

Issue #454, opened 2016-09-12
Prepare Measures/Objectives/Decisions for EPS 


arnaud-m:
Some fixes, improvements and modifications coming from the implementation of Embarrassingly Parallel Search for Choco.

arnaud-m:
A few questions arise while refactoring the objective manager.

What is the expected behavior of the SAT manager ? Actually it either [returns null](https://github.com/chocoteam/choco-solver/blob/master/src/main/java/org/chocosolver/solver/objective/BoundsManager.java#L171) or [raises an exception](https://github.com/chocoteam/choco-solver/blob/master/src/main/java/org/chocosolver/solver/objective/BoundsManager.java#L107).

I doubt that it is required to [check for null](https://github.com/chocoteam/choco-solver/blob/master/src/main/java/org/chocosolver/solver/objective/ObjectiveManager.java#L225) when updating the bounds.
Indeed, they are initialized in the constructor before the object becomes accessible.
I wonder if these lines do not come from older version.

cprudhom:
It is designed to be the default implementation for satisfaction problem, so there is need to retrieve the best bound.
We wanted to define a unique entry point for the two major resolution types (CSP and COP).
This is not a class directly manipulated by users, that may explain why it became ugly :)

We are open to suggestions/modifications, but please avoid too much abstraction.

PR #455, opened 2016-09-12
Modifications of measures #454

arnaud-m:
Just trying a small pull request ;)


src/main/java/org/chocosolver/solver/search/measure/MeasuresRecorder.java
```diff
@@ -29,413 +29,177 @@
...
-
-    /**
-     * Search state
-     */
-    protected SearchState state;
-
-    //***********************************************************************************
-    // CONSTRUCTOR
-    //***********************************************************************************
+    private long startingTime;
```
arnaud-m: No, it is a private field used in a lambda function.

chocotrust:
I'll add it manually. I may be keep StopWatch name instead of Timer, i'll see

chocotrust:
It seems that you have an indentation problem. 
I will soon provide the file format we use.

arnaud-m:
I tried to modify the "Java Convention" built-in formatter of Eclipse to match yours.
I hope that export will work (see [here](http://stackoverflow.com/questions/36432438/importing-code-style-formatting-settings-into-eclipse-from-intellij-idea)).

PR #462, opened 2016-09-16
Refactor the objective managers

arnaud-m:
The actual code based on generics is messy and there are duplicates between the bounds and objective managers. 

Here is a polymorphic approach that could replace the generics approach.


src/main/java/org/chocosolver/solver/objective/ObjectiveFactory.java
```diff
@@ -0,0 +1,211 @@
...
+        super();
+    }
+
+    /**
+     * Define a manager for satisfaction problems.
+     * 
+     * @return a singleton object
+     * 
+     */
+    public static IObjectiveManager<Variable> SAT() {
```
arnaud-m: I keep the old method name, but I would not mind to change it.

PR #467, opened 2016-11-02
measures, decisions, and objective managers are serializable #454

arnaud-m:
introduce a factory of serializable decisions


arnaud-m:
The fields mentioned by codacy are useless and could be removed or deprecated, because they only ensure backward-compatibility.

cprudhom:
I'll merge the PR asap and take your remarks into account

src/main/java/org/chocosolver/solver/search/strategy/assignments/DecisionOperator.java
```diff
@@ -63,175 +65,14 @@
...
-            return var.getLB() <= value;
-        }
-
-        @Override
-        public DecisionOperator opposite() {
-            return int_split;
-        }
-    };
-
+    DecisionOperator<IntVar> int_eq = DecisionOperatorFactory.makeIntEq();
```
cprudhom: We should consider calling directly the factory and remove these methods from DecisionOperator

cprudhom:
I let you remove useless fields from DecisionOperator, then tell me and I'll merge the PR.

arnaud-m:
OK, I will remove the fields and commit today.

gregy4:
Removed field are not useless, the code is in release 4.0.1 and not mentioned in release notes. I already talked about it with Charles, deprecated directive should be used. It is waste of time for others to find out why their code cannot be compiled.

PR #468, opened 2016-11-24
Fixed ArrayUtils.reverse() for generic data types when the array size is even

rhelmke:
The generic ArrayUtils.reverse() calculates the middle-element for Arrays containing an even number of elements wrong. Arrays containing an even amount of elements are being reversed improperly. This pull-Request fixes that issue.


cprudhom:
Hi Rene, 
And thank you for the PR.
It seems that a test is not working anymore, `SolverTest.test1HBFS` so I can't accept your PR.
But I'll have a look soon.

PR #493, opened 2017-01-17
optimize first fail variable selector

arnaud-m:
- Before: at least 3 tests per variable.
- After: at most 3 tests per variable.

### DIFF
--- a/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/FirstFail.java
+++ b/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/FirstFail.java
@@ -22,7 +22,7 @@ import org.chocosolver.solver.variables.IntVar;
  */
 public class FirstFail implements VariableSelector<IntVar>, VariableEvaluator<IntVar> {
 
-    private IStateInt lastIdx; // index of the last non-instantiated variable
+    private final IStateInt lastIdx; // index of the last non-instantiated variable
 
     /**
      * <b>First fail</b> variable selector.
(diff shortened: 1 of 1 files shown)


arnaud-m:
Eclipse raises a compilation error in `org.chocosolver.solver.search.strategy.Search.bestBound(AbstractStrategy)` because the class `BoundSearch` class is not found.

src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/FirstFail.java
```diff
@@ -31,33 +31,36 @@
...
+        while(idx < variables.length && variables[idx].isInstantiated()) {
+            idx++;
+        }
+        lastIdx.set(idx);
+        //search for the leftmost variable with smallest domain
+        while(idx < variables.length) {
+            final int dsize = variables[idx].getDomainSize();
+            if (dsize < smallDSize) {
+             // the variable is candidate for having the smallest domain
+                if( dsize > 1 ) {
```
arnaud-m: I will do it as soon as the Eclipse compilation error is fixed.

arnaud-m:
```
Failed tests: 
ObjectiveTest.test1:78->max:120 expected [2] but found [21]
```

Apparently, the test fails because the search selects the smallest value of the domain rather than the largest.

src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/FirstFail.java
arnaud-m: and the testing error.

cprudhom:
Yes, that's due to an earlier commit.
I'll handle them all.

arnaud-m:
Ok, I am waiting for your update.

cprudhom:
Now, you can update your code to deal with codacy advices.
Make sure you rebase your branch first:
```bash
$ git fetch
$ git rebase master
```

PR #501, opened 2017-01-28
Correct the explanation to be produced by PropCount

mwahbi:
A simple comparison with PropCountVar can show that the previous explainer of PropCount is not accurate. I can provide a simple example if necessary!


cprudhom:
Why did you delete it ? Is it incorrect ?

mwahbi:
It looks that what I proposed doesn't solve my issue, I still need to investigate why it didn't, now I am using PropCountVar. Anyway the current version seems to not work PropCount_AC! 
I will let you know if I found why and what is missing?

PR #540, opened 2017-06-29
making an accurate implementation of PGLNS from Perron et al. CP2004 & updating toStrings

mwahbi:
I think the PGLNS [Perron et al., CP2004] was not accurate, this is an accurate proposition of its implementation.

### DIFF
--- a/src/main/java/org/chocosolver/solver/Model.java
+++ b/src/main/java/org/chocosolver/solver/Model.java
@@ -25,12 +25,14 @@ import org.chocosolver.solver.objective.ObjectiveFactory;
 import org.chocosolver.solver.propagation.IPropagationEngine;
 import org.chocosolver.solver.propagation.NoPropagationEngine;
 import org.chocosolver.solver.propagation.PropagationTrigger;
+import org.chocosolver.solver.search.strategy.decision.IntDecision;
 import org.chocosolver.solver.variables.*;
 
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * The <code>Model</code> is the header component of Constraint Programming.
@@ -65,6 +67,8 @@ public class Model implements IModel {
     public static final String MINISAT_HOOK_NAME = "H_MINISAT";
 
     public static final String NOGOODS_HOOK_NAME = "H_NOGOODS";
+
+    public static final String SOLVER_HOOK_NAME = "H_SOLVER";
 
     /**
      * Settings to use with this solver
@@ -117,11 +121,6 @@ public class Model implements IModel {
      * Environment, based of the search tree (trailing or copying)
      */
     private final IEnvironment environment;
-
-    /**
-     * Resolver of the model, controls propagation and search
-     */
-    private final Solver solver;
 
     /**
      * Variable to optimize, possibly null.
@@ -190,7 +189,6 @@ public class Model implements IModel {
         this.cachedConstants = new TIntObjectHashMap<>(16, 1.5f, Integer.MAX_VALUE);
         this.objective = null;
         this.hooks = new HashMap<>();
-        this.solver = new Solver(this);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


cprudhom:
Could you explain a little bit the modification your brought ?

mwahbi:
Hi @cprudhom,
The story is when updating the list of candidates. In the previous implementation the candidate list are not sorted, while the paper says the candidate should be sorted by the impact on the search tree. In the previous implementation candidate is a SortedMap sorted by the keys (vars index). So, one solution is to create a comparator based on the values or just use and array as I proposed. 
I hope it is clear! 
Regards

cprudhom:
Ok, you're right, my bad. `SortedMap` killed me once again :/
I'll push your patch asap.
Thank you for your contribution once again !

cprudhom:
I certainly won't integrate `Solver.toString()` as is, since it is a copy/paste of `Model.toString()` except for decision path.
Where do you call this method ?

cprudhom:
Plus, did you modify RPGLNS too ? Because it does not appear in the patch.

mwahbi:
I can see my modifications on PropagationGuidedNeighborhood, I have 4 files changed, can you see what I proposed as implementation (get rid of SortedMap and use arrays)?

cprudhom:
Yes, I also received 4 files, but `ReversePropagationGuidedNeighborhood` was not in and it extends `PropagationGuidedNeighborhood`.

I updated its code as follow:
```java
@Override
    protected void update(DecisionPath decisionPath) throws ContradictionException {
        while (logSum > fgmtSize && fragment.cardinality() > 0) {
            // 1. pick a variable
            int id = selectVariable();

            // 2. fix it to its solution value
            if (vars[id].contains(bestSolution[id])) {  // to deal with objective variable and related

                mModel.getEnvironment().worldPush();
                vars[id].instantiateTo(bestSolution[id], Cause.Null);
                mModel.getSolver().propagate();
                fragment.clear(id);

                for (int i = 0; i < n; i++) {
                    int ds = vars[i].getDomainSize();
                    if (fragment.get(i)) { // if not frozen until now
                        if (ds == 1) { // if fixed by side effect
                            fragment.clear(i); // set it has fixed
                        } else {
                            int closeness = (int) ((dsize[i] - ds) / (dsize[i] * 1.) * 100);
                            //                            System.out.printf("%d -> %d :%d\n", dsize[i], ds, closeness);
                            if (closeness > 0) {
                                all[i] = closeness; // add it to candidate list
                            }
                        }
                    }
                }
                mModel.getEnvironment().worldPop();
                candidates = IntStream.range(0, n)
                        .filter(i -> fragment.get(i) || all[i] == 0)
                        .boxed()
                        .sorted(Comparator.comparingInt(i -> all[i]))
                        .limit(listSize)
                        .collect(Collectors.toList());
                logSum = 0;
                for (int i = fragment.nextSetBit(0); i > -1 && i < n; i = fragment.nextSetBit(i + 1)) {
                    logSum += Math.log(vars[i].getDomainSize());
                }
            } else {
                fragment.clear(id);
            }

        }
        for (int i = fragment.nextSetBit(0); i > -1 && i < n; i = fragment.nextSetBit(i + 1)) {
            if (vars[i].contains(bestSolution[i])) {
                impose(i, decisionPath);
            }
        }
        mModel.getSolver().propagate();

        logSum = 0;
        for (int i = 0; i < n; i++) {
            logSum += Math.log(vars[i].getDomainSize());
        }
    }

```
If you agree, I push this version.

mwahbi:
Ouuups I didn't change the reverse, I think your changes on reverse are accurate, go on ;)

cprudhom:
Is it the right PR ?

cprudhom:
I mean, we should try to separate the purpose.
I will ignore the last push of the PR and wait until you create a new one with only the modifications related to #541

cprudhom:
I integrated the PR in the master branch.
Feel free to add a new PR related to #541.

cprudhom:
Well, I should be able to get your last commit

mwahbi:
@cprudhom I am not an expert in github, so I don't know how to fix it. You can ignore the my commit untill I figured out how to do it properly, any advise?

Issue #541, opened 2017-06-29
solver instance


mwahbi:
Hi @cprudhom, @jgFages, 
I think the solver should not be final in the model. If somebody wants to extend the solver and to maneuver its extensions. Why this? 
I have a user communicating with the solver through (1) adding decisions and (2) adding nogoods or reporting nogoods when the problem is infeasible regarding (1). But, I have restrictions on the processing of doing so: (3) a solver is not allowed to change the decisions in (1) and (4) can only make decision on a subset of variables. 
For doing (4) it is easy defining a strategy on the subset of variables is the solution. 
I think for doing (3) there two options: The first adding new equality constraints (maybe it will be slow) but it will be a problem when we can not find a solution and we need to report the nogood to the user (2), i.e., a subset of (1). The second one is to manipulate the decision path adding new decisions from (1) to it and in this case as well we need to change the repair phase because we cannot refute a decision form (1). 
I think all those issues can be solved if the solver/model is modular and will avoid to change the source codes and use directly the maven dependency. Moreover, it will ease maintaining the codes. Right? 
Any other thoughts on my issues are very welcome!
Cheers,

cprudhom:
I agree that it could be useful when one has its own implementation of `Solver`.
But I think we should avoid using hook to store such crucial object.
Indeed, `model.getSolver()` is used almost everywhere.

I suggest we define a method in `Model` named `createSolver()` with no parameter which, by default, create a `Solver` object as for now.
This method can be overridden  by anyone who wants to change the way a solver is created.

Alternatively, we could this new method in `Settings` instead of `Model`.

mwahbi:
I initially propose createSolver()/setSolver() but change it later to the hook thinking that would be better. But, now I see the issue of hook so yes it shouldn't be a hook. I am for createSolver()/setSolver(). So, it is up to you.

jgFages:
Hi,
First of all, I am not sure I really understood what you need but it seems that:
You have a server with a choco model. 
You receive a user request that applies a few constraints/decision
You want a solution or an explanation of failure

1) -> You can do all this without changing anything in source code. 
Applying constraints should be ok but actually you could instantiate it directly manually if it is only assignments (I do it on some configurator projects).
2) -> What do you use to explain the conflict? Maybe here is the issue

mwahbi:
Hi JG,
[...]

PR #542, opened 2017-06-29
For resolving #541, allowing the user to change the solver instance

mwahbi:
I propose setSolver

### DIFF
--- a/src/main/java/org/chocosolver/solver/Model.java
+++ b/src/main/java/org/chocosolver/solver/Model.java
@@ -121,7 +121,7 @@ public class Model implements IModel {
     /**
      * Resolver of the model, controls propagation and search
      */
-    private final Solver solver;
+    private Solver solver;
 
     /**
      * Variable to optimize, possibly null.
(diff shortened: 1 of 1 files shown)


mwahbi:
Hi Charles, 
So, the final decision is keep the solver final in model. Right? :(

cprudhom:
Exactly. If you inherit from `Model` I can add a protected or package-private constructor with `Solver` in it, but no setter

mwahbi:
@cprudhom inherit from <code>Model</code> will do the job I want, so if that solution suits you it is good for what I want to do. Thanks you!

cprudhom:
Ok, here is how you should do, this deliberately complicate.

```java
// 1. Declare your own Settings which overrides 'initSolver(Model m)'
Settings s = new Settings() {
    @Override
    public Solver initSolver(Model model) {
        return new MySolver(model);
    }
}; 
// Then create the model with the following constructor:
Model model = new Model(new EnvironmentBuilder().fromFlat().build(), "my name", s);
```

Maybe we could consider creating a model builder (but I'm not a big fan of this).

That will be commit soon and part of the next release

PR #576, opened 2017-12-04
Added minor patch for missing initialization when calling propagate() before solve()

RobertHilbrich:
I would like to propose this pull request. In ASSIST, we are calling propagate after adding a set of constraints to check, if the problem is still solvable. Unfortunately, initialize() is only called as part of solve(), so our propagate() calls hit the assertion. We could fix this by lines:

[ASSISTSolver.xtend](https://github.com/RobertHilbrich/assist-public/blob/master/ch.hilbri.assist.mapping/src/ch/hilbri/assist/mapping/solver/AssistSolver.xtend) lines 90-92, but this is not good for readability.

@chocoteam/core-developer


cprudhom:
Hi, 

Thank you for the PR. I think the fix should be extended to PropSat.
I will push it as soon as possible.

best

PR #598, opened 2018-10-15
Add an alternative constructor to AbstractVariable (explicit EvtScheduler) to maintain choco-graph compatibility

dimitri-justeau:
Some modifications since choco 4.0.6 broke the compatibility with choco-graph, but... I love both choco and choco-graph and I want to use both newest versions :cry:

The problem is located in the constructor of AbstractVariable. In fact, it tries to do operations on the EvtScheduler, which is not yet instantiated in choco-graph, since java reflexion is used afterwards for it.

As a workaround, I added an alternative constructor which accepts an EvtScheduler as an argument.

However, I believe that this could be handled more elegantly, such as for instance by defining the typeAndKind() and the EvtScheduler implementation at the class level. I did not want to make any design choice, but I am ok to work on this if needed.

PS: I have a commit for the choco-graph side, that I will pull request if this one is accepted, and then all the world will use choco and choco-graph newest versions, yeeeahhh !

### DIFF
--- a/src/main/java/org/chocosolver/solver/variables/impl/AbstractVariable.java
+++ b/src/main/java/org/chocosolver/solver/variables/impl/AbstractVariable.java
@@ -149,6 +149,27 @@ public abstract class AbstractVariable implements Variable {
     //////////////////////////////////////////////////////////////////////////////////////
 
     /**
+     * Constructor to be used by choco extension. This (for the moment) a workaround for ensuring choco-graph
+     * compatibility (cf. comment in second constructor).
+     * @param name
+     * @param model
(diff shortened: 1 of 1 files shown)


cprudhom:
Thank you for noticing the regression.
My suggestion is to define a method named `createScheduler()` either in `Variable` or in `AbstractVariable` (as abstract) and to define it concretely in all implementations.
The method will still be called in the constructor.

Does that fit your needs ?

dimitri-justeau:
Yes ! Should I close this PR and implement your suggestion ?

cprudhom:
Yes please

PR #613, opened 2019-01-18
Ineighbor

### LINKED ISSUE #538: Explanations and LNS may create strange decisions
Explanations: decisions var=val where val does not belong to var domain
(ExplanationEngineTest.testLSsmall)
LNS: decisions var = val where var is already equal to val
(SolverTest.test6)

This is not a big problem but it would be more convenient to avoid it

jgFages:
Fixes #538 on other variable kinds (sets and reals) .

Changes proposed in this PR:
- Put back INeighbor to define LNS over different types of variables
- Built-in Random LNS for a SetVar
- Rename Neighbor into IntNeighbor
- More tests
 
@chocoteam/core-developer


alberthendriks:
I don't understand why that ^ circleci is failing? The end of the logs says:
> [INFO] Building jar: /home/circleci/project/target/choco-solver-4.10.1-SNAPSHOT.jar
> Received 'killed' signal

The same code with the same git hash is succeeding on my fresh circleci account: https://circleci.com/gh/alberthendriks/choco-solver/3

Maybe someone try "rerun" in circleci?

jgFages:
it does not like to run the same test both on travis and circleci (this is why I had excluded 1s and 10s from travis)
actually we can remove codacy completely, travis is enough

PR #620, opened 2019-03-29
Fixes issue on Modulo constraint

### LINKED ISSUE #555: Modulo
current modulo constraint has all arguments variables: x1%x2 = x3
I believe it is very rare that x2 is a variable

- a specific implementation with x2 as a fixed int should improve runtime

- Having another signature with both x2 and x3 fixed would also make sense, and it can be implemented easily as a member over a precomputed array

No hurry but to keep in mind

ArthurGodet:
Fixes #555 .

Changes proposed in this PR:
- Add 3 different constraints for Modulo constraint on 1, 2 or 3 variables
- Add corresponding tests
- Update IIntConstraintFactory to take these changes into account

The propagators for 2 and 3 variables (a.k.a PropModXY and PropModXYZ) could probably be improved by reacting to fine events
 
@chocoteam/core-developer


cprudhom:
Thank you for the PR. 
Do you have any idea of the performance differences wrt the previous version ?

CP

cprudhom:
In addition, why did you close it ?

ArthurGodet:
I closed it because I wanted to change how to manage the case where the modulo is 0, and it introduced some bugs, unknown until then. And I didn't want to add too many commits for "just bugs correcting".

However, I am still working on this and will open a new PR when everything is clean and without bugs. In terms of performance, for what I have tested, I get a gain of at least 50% in resolution time, and a diminution of up to 10% in the number of nodes.

PR #621, opened 2019-04-03
Modulo propagators and better formulation

### LINKED ISSUE #555: Modulo

ArthurGodet:
Fixes #555  .

Changes proposed in this PR:
- Treat differently the modulo constraint declaration to authorize int parameters (IIntConstraintFactory)
- Add propagators for modulo constraints for 2 and 3 variables (PropModXY and PropModXYZ)
- Add tests for the two propagators as well as for the formulation with table constraints

I have tested these changes on home-made problems. In terms of performance, this new approach use up to 50% variables and constraints, is about twice faster and filters better (up to 10% nodes less) than the old approach.
 
@chocoteam/core-developer


ArthurGodet:
I think that the 0.02% of code coverage loss is due to the fact that my code does not use model.div(...), model.times(...) and model.sum(...).

PR #636, opened 2019-10-10
Add Neighbourhood Search

ingambe:
# Introduction

This pull request adds a new search strategy which allows us to try first some initial values in the value selection if these values are still present in the domain of the variables, otherwise we select the value using the value selector.
This way we can try to find a solution close to a previous assignment. This can be useful in case of recovery (a solution became infeasible because of a new constraint) or if we have some information about possible solutions (for example, we know that often certain values are assigned to specific values), etc.

# How to use it ?

To create the Neighbourhood Search, you need to use the static method [Search@intNeighbourhoodSearch](https://github.com/ingambe/choco-solver/blob/master/src/main/java/org/chocosolver/solver/search/strategy/Search.java#L420)
This method takes as an input, as usual, the variables the search needs to focus on, and a list of variable for which we need to try to apply an initial value, the initial values and the strategy to use for variable selection and value selection in case we failed to assign the initial value.

# Code location

## Code

-[IntNeighbourhood](https://github.com/ingambe/choco-solver/blob/master/src/main/java/org/chocosolver/solver/search/strategy/selectors/values/IntNeighbourhood.java): the wrapper which keep the initial value and try to assign if possible

## Tests

-[NeighbourhoodSearchTest](https://github.com/ingambe/choco-solver/blob/master/src/test/java/org/chocosolver/solver/search/strategy/strategy/NeighbourhoodSearchTest.java): tests to check the selection mechanism of the initial value.
-[IntNeighbourhoodTest](https://github.com/ingambe/choco-solver/blob/master/src/test/java/org/chocosolver/solver/search/strategy/selectors/values/IntNeighbourhoodTest.java): tests for the value selector, check if we select correctly the value if it is inside the domain, don't select the initial value if not in the domain, and use the value selector of the search strategy if no initial value or not in the domain.

Changes proposed in this PR:
- A new search strategy, which is a wrapper for another search strategy
 
@chocoteam/core-developer

### DIFF
--- a/src/main/java/org/chocosolver/solver/search/strategy/Search.java
+++ b/src/main/java/org/chocosolver/solver/search/strategy/Search.java
@@ -407,12 +407,12 @@ public class Search {
     }
 
     /**
-     *
-     * @param vars
-     * @param selected
-     * @param initial
-     * @param strategy
-     * @return
+     * Neighbourhood Search strategy
+     * @param vars variable concerned by the search
+     * @param selected variable for which the value need to be reused (if possible)
+     * @param initial the value we need to try to assign if the variable is selected
+     * @param strategy the strategy to apply if we can/don't select the variable
+     * @return the strategy
      */
     public static AbstractStrategy<IntVar> intNeighbourhoodSearch(IntVar[] vars, boolean[] selected, int[] initial, IntStrategy strategy) {
         List<Integer> initialValues = new LinkedList<>();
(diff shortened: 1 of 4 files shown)


cprudhom:
First of all, thank you for your contribution !
I have a question : can you tell me what differs from `IntDomainLast` which is based on solution object (the last one found or one given as input) and tend to branch first on values as long as they appear in current variables' domain ?

ingambe:
@cprudhom Oh! Sorry, I was not aware that such Search strategy existed previously.
It's exactly the same, sorry about that.

PR #649, opened 2019-12-26
solver time stop criterion returns true despite search not started after call to hardReset()

### LINKED ISSUE #648: [BUG] Adding stop criterion after `Solver.hardReset()` failure
Hi,

in my model I have several consecutive solving phases, where two such phases are separated by a `Solver.hardReset()` and a (timeout) `Solver.addStopCriterion()`. I have observed the behavior that, directly after adding the timeout stop criterion, the solver stops because of `Solver.isStopCriterionMet()`. This is directly after setting a rather long timeout, so I would not except this behavior.

This was also discussed as question 1 in https://groups.google.com/d/msg/choco-solver/xrBhq2RKVZM/gY-Uu700CgAJ

Many thanks in advance!

Magnus

## Expected behavior

The solver should continue the given time set by the timeout stop criterion.

## Actual behavior

The solver directly stops after setting the timeout stop criterion.

## Possible Solution

A reflection and what I currently use as workaround is adding a `Solver.getMeasures().reset()` directly after `Solver.hardReset()`. I would expect there is some problem with the time counter in `MeasuresRecorder`.

## Steps to Reproduce (for bugs)

```
    @Test
    public void testStopCriterion() {
        final Model model = makeNonTrivialModel();

        final Solver solver = model.getSolver();

        final long timeLimit = 2 * 1000000000L;

        solver.addStopCriterion(new TimeCounter(model, timeLimit));
        Assert.assertFalse(solver.isStopCriterionMet());
        solver.solve();

        solver.hardReset();

        solver.addStopCriterion(new TimeCounter(model, timeLimit));
        Assert.assertFalse(solver.isStopCriterionMet());
        solver.solve();
    }


    // Slightly adapted Golomb Ruler instance from:
    // https://github.com/chocoteam/samples/blob/master/src/main/java/org/chocosolver/samples/integer/GolombRuler.java
    //
    // The choice of model is irrelevant to this test, but must not be trivial to solve.
    private Model makeNonTrivialModel() {
        int m = 12;
        Model model = new Model("GolombRuler");
        IntVar[] ticks = model.intVarArray("a", m, 0, (1 << m + 1) - 1, false);

        model.arithm(ticks[0], "=", 0).post();

        for (int i = 0; i < m - 1; i++) {
            model.arithm(ticks[i + 1], ">", ticks[i]).post();
        }

        IntVar[] diffs = model.intVarArray("d", (m * m - m) / 2, 0, (1 << m + 1) - 1, false);
        IntVar[][] m_diffs = new IntVar[m][m];
        for (int k = 0, i = 0; i < m - 1; i++) {
            for (int j = i + 1; j < m; j++, k++) {
                model.scalar(new IntVar[]{ticks[j], ticks[i]}, new int[]{1, -1}, "=", diffs[k]).post();
                model.arithm(diffs[k], ">=", (j - i) * (j - i + 1) / 2).post();
                model.arithm(diffs[k], "-", ticks[m - 1], "<=", -((m - 1 - j + i) * (m - j + i)) / 2).post();
                model.arithm(diffs[k], "<=", ticks[m - 1], "-", ((m - 1 - j + i) * (m - j + i)) / 2).post();
                m_diffs[i][j] = diffs[k];
            }
        }
        model.allDifferent(diffs, "BC").post();
        return model;
    }
```

## Environment

* Choco-solver version: 4.10.2
* Java version: openjdk version "11.0.1" 2018-10-16

ArthurGodet:
Fixes #648 .

Changes proposed in this PR:
- Adds an `if` test in `Solver.isStopCriterionMet()` to avoid checking time stop criterion if the search is not started (this behaviour can happen after `Solver.hardReset()` )
 
@chocoteam/core-developer

### DIFF
--- a/src/main/java/org/chocosolver/solver/search/measure/MeasuresRecorder.java
+++ b/src/main/java/org/chocosolver/solver/search/measure/MeasuresRecorder.java
@@ -98,8 +98,8 @@ public final class MeasuresRecorder extends Measures {
         state = SearchState.NEW;
         objectiveOptimal = false;
         solutionCount = 0;
-        timeCount = 0;
         stopStopwatch();
+        timeCount = 0;
         nodeCount = 0;
         backtrackCount = 0;
         failCount = 0;
--- a/src/main/java/org/chocosolver/solver/Solver.java
+++ b/src/main/java/org/chocosolver/solver/Solver.java
@@ -20,7 +20,6 @@ import org.chocosolver.solver.objective.ObjectiveFactory;
 import org.chocosolver.solver.propagation.PropagationEngine;
 import org.chocosolver.solver.search.SearchState;
 import org.chocosolver.solver.search.limits.ICounter;
-import org.chocosolver.solver.search.limits.TimeCounter;
 import org.chocosolver.solver.search.loop.Reporting;
 import org.chocosolver.solver.search.loop.learn.Learn;
 import org.chocosolver.solver.search.loop.learn.LearnNothing;
--- a/src/main/java/org/chocosolver/solver/Solver.java
+++ b/src/main/java/org/chocosolver/solver/Solver.java
@@ -752,9 +752,6 @@ public class Solver implements ISolver, IMeasures, IOutputFactory {
     public boolean isStopCriterionMet() {
         boolean ismet = false;
         for (int i = 0; i < criteria.size() && !ismet; i++) {
-            if(criteria.get(i) instanceof TimeCounter && this.getNodeCount() <= 0) {
-                continue; // time limit for search is not reached when search is not started
-            }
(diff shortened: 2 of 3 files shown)


src/main/java/org/chocosolver/solver/search/measure/MeasuresRecorder.java
```diff
@@ -98,8 +98,8 @@ public void reset() {
         state = SearchState.NEW;
         objectiveOptimal = false;
         solutionCount = 0;
-        timeCount = 0;
         stopStopwatch();
+        timeCount = 0;
```
mr-tl: It could be good to add a comment that it is essential that the assignment `timeCount = 0` is after the call to `stopStopwatch()`.

PR #663, opened 2020-01-21
Improving PropDiffN performance

ArthurGodet:
Fixes # .

Changes proposed in this PR:
- Improving PropDiffN performance by changing some data structure and slightly improving energy check
 
@chocoteam/core-developer


solver/src/main/java/org/chocosolver/solver/constraints/nary/PropDiffN.java
```diff
@@ -36,95 +35,117 @@
...
-                for (int j = i + 1; j < n; j++) {
-                    if (mayOverlap(i, j)) {
-                        overlappingBoxes.addEdge(i, j);
-                        if (boxInstantiated(i) && boxInstantiated(j)) {
-                            fails(); // TODO: could be more precise, for explanation purpose
+        boolean hasFiltered = true;
+        while(hasFiltered) {
+            hasFiltered = false;
+            if(PropagatorEventType.isFullPropagation(evtmask)) {
+                boxesToCompute.clear();
```
cprudhom: I suppose `boxesToCompute.resetQuick()` is enough in that case.

solver/src/main/java/org/chocosolver/solver/constraints/nary/PropDiffN.java
```diff
@@ -36,95 +35,117 @@
...
+                for (int i = 0; i < n; i++) {
+                    boxesToCompute.add(i);
+                    for (int j = i + 1; j < n; j++) {
+                        if (mayOverlap(i, j)) {
+                            overlappingBoxes.addEdge(i, j);
+                            if (boxInstantiated(i) && boxInstantiated(j)) {
+                                fails(); // TODO: could be more precise, for explanation purpose
+                            }
+                        } else {
+                            overlappingBoxes.removeEdge(i, j);
```
cprudhom: Isn't that too strong ? There are cases where `mayOverlap` returns `false` and only one of the two conditions is falsified ?

solver/src/main/java/org/chocosolver/solver/constraints/nary/PropDiffN.java
ArthurGodet: I suppose it is, yes.

solver/src/main/java/org/chocosolver/solver/constraints/nary/PropDiffN.java
ArthurGodet: An edge in `overlappingBoxes` indicates that rectangle i and rectangle j might overlap in future instantiations and filterings, which should not happen.

ArthurGodet:
The gain on pure random instances was about 20% faster execution for the same number of nodes for the same search strategy (inputOrder).

PR #666, opened 2020-01-24
Add a pure Java alternative to Ibex (to be completed though).

cprudhom:
Many operators are not supported yet (neg, abs, min, max, pow, atan2, exp, ln, sqrt, tan, acos, asin, atan, sinh, tanh, acosh, atanh), so some tests fail. Any contribution will be very much appreciated !
Now, the default behavior is pure Java.
Note that, RealView are not supported in PropEquation and a proper constraint should be posted in addition.

- Update continuous arithmetical expression to be inputed in `CEquation` using HC4 algorithm for filtering value
- Add syntaxic sugar to ease equation extraction.
- Add eq(RealVar,IntVar) constraint (to be used instead of views)
- Add element(RealVar, double[], IntVar) constraint


solver/src/main/java/org/chocosolver/solver/expression/continuous/relational/BiCReExpression.java
```diff
@@ -73,6 +75,14 @@ public Constraint ibex(double p) {
     }
 
 
+    @Override
+    public Constraint equation() {
+        CArExpression expression = e1.sub(e2);
```
ArthurGodet: Is it always `sub`, whatever the mathematical operator ?

solver/src/main/java/org/chocosolver/solver/expression/continuous/relational/PropEquation.java
```diff
@@ -0,0 +1,234 @@
...
+                    var.silentlyAssign(unexplored[--idx]);
+                    depth = depths[idx];
+                } else {
+                    fin = true;
+                }
+            }
+        }
+
+        // Propagation
+        var.silentlyAssign(oldValue[0], oldValue[1]);
```
ArthurGodet: Should be removed

solver/src/main/java/org/chocosolver/solver/expression/continuous/relational/BiCReExpression.java
cprudhom: Yes it is. Some improvements may be brought at the expense of straightforwardness I suppose.

solver/src/main/java/org/chocosolver/solver/expression/continuous/relational/PropEquation.java
cprudhom: No it shouldn’t since this call ensures that the initial values are restored. Indeed, the two previous while-loop change these bounds abruptly to detect inconsistency.

solver/src/main/java/org/chocosolver/solver/expression/continuous/arithmetic/BiCArExpression.java
```diff
@@ -121,6 +132,146 @@ public RealVar realVar(double p) {
...
+            wox.add(this);
+        }
+        return dependsOnX;
+    }
+
+    @Override
+    public void init() {
+        if(l == null && u == null) {
+            l = model.getEnvironment().makeFloat(Double.NEGATIVE_INFINITY);
+            u = model.getEnvironment().makeFloat(Double.POSITIVE_INFINITY);
```
ArthurGodet: Might it be preferable to use +/- Double.MAX_VALUE ?

PR #669, opened 2020-02-11
Enable nogood stealing (on restart) for ParallelPortfolio

cprudhom:
Offer the possibility, when using a `ParallelPortfolio`, to steal no-goods from decision path of other models. The unique (strong) assumption is that all models are *equivalent* :
- each variable exists in all models, 
- each variable has the same ID in all models 
- each variable is declared at the same position in all models. 

When this option is on, any worker with a restart policy will extract a nogood on each restart, based on its decision path, **and** based on current decision path of other models in the portfolio.

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/search/loop/monitors/NogoodFromRestarts.java
+++ b/solver/src/main/java/org/chocosolver/solver/search/loop/monitors/NogoodFromRestarts.java
@@ -9,16 +9,21 @@
  */
 package org.chocosolver.solver.search.loop.monitors;
 
-import org.chocosolver.sat.SatSolver;
-import org.chocosolver.solver.Model;
-import org.chocosolver.solver.constraints.nary.sat.PropNogoods;
-import org.chocosolver.solver.search.strategy.assignments.DecisionOperatorFactory;
-import org.chocosolver.solver.search.strategy.decision.Decision;
-import org.chocosolver.solver.search.strategy.decision.IntDecision;
-import org.chocosolver.solver.search.strategy.decision.SetDecision;
-
 import java.util.ArrayDeque;
 import java.util.Arrays;
+import org.chocosolver.sat.SatSolver;
+import org.chocosolver.solver.Model;
+import org.chocosolver.solver.constraints.nary.sat.NogoodStealer;
+import org.chocosolver.solver.constraints.nary.sat.PropNogoods;
+import org.chocosolver.solver.search.strategy.assignments.DecisionOperator;
+import org.chocosolver.solver.search.strategy.assignments.DecisionOperatorFactory;
+import org.chocosolver.solver.search.strategy.decision.Decision;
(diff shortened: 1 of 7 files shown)


solver/src/main/java/org/chocosolver/solver/search/loop/monitors/NogoodFromRestarts.java
```diff
@@ -46,107 +52,110 @@
...
+            return asLit(
+                nogoodStealer.getById(id.getDecisionVariable(), png.getModel()), id.getDecOp(),
+                id.getDecisionValue()
+            );
+        } else if (decision instanceof SetDecision) {
+            SetDecision id = (SetDecision) decision;
+            return asLit(
+                nogoodStealer.getById(id.getDecisionVariable(), png.getModel()), id.getDecOp(),
+                id.getDecisionValue()
+            );
```
ArthurGodet: This code could be easier by transfering `getDecOp()` to the `Decision<E>` class, where it would return a `DecisionOperator<E>`.

solver/src/main/java/org/chocosolver/solver/search/loop/monitors/NogoodFromRestarts.java
```diff
@@ -46,107 +52,110 @@
...
+            throw new UnsupportedOperationException("Cannot deal with such operator: " + op);
+        }
+        return l;
+    }
+
+    private int asLit(SetVar var, DecisionOperator<SetVar> op, int val) {
+        int l;
+        if (op == DecisionOperatorFactory.makeSetForce()) {
+            l = SatSolver.negated(png.Literal(var, val, true));
+        } else if (op == DecisionOperatorFactory.makeSetRemove()) {
```
ArthurGodet: Method `equals()` should be used instead of `==`.

solver/src/main/java/org/chocosolver/solver/search/loop/monitors/NogoodFromRestarts.java
cprudhom: I'm not sure this would be possible considering that `RealDecision` does not rely on `DecisionOperator` since it requires `value` to be a `double` and not an `int` as needed for `IntDecision` and `SetDecision`.
The same goes for `RootDecision`, but that's actually not a problem.

solver/src/main/java/org/chocosolver/solver/search/loop/monitors/NogoodFromRestarts.java
cprudhom: U're right

solver/src/main/java/org/chocosolver/solver/search/loop/monitors/NogoodFromRestarts.java
cprudhom: This should be evaluated in another PR, because I'm confident with the performance when `DecisionOperator` is generified (and its consequence on `Decision`, `Search` and strategies)

PR #674, opened 2020-03-16
Introduce new propagator for nValues constraint based on witness/watch

ArthurGodet:
Changes proposed in this PR:
- Global improvements on the performance of the nValues constraint
 
@chocoteam/core-developer


solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
```diff
@@ -0,0 +1,164 @@
...
+import org.chocosolver.util.objects.setDataStructures.ISet;
+import org.chocosolver.util.objects.setDataStructures.SetFactory;
+import org.chocosolver.util.objects.setDataStructures.SetType;
+import org.chocosolver.util.tools.ArrayUtils;
+
+public class PropNValue extends Propagator<IntVar> {
+    private IntVar nValue;
+    private final int n;
+    private int[] concernedValues;
+    private IStateInt[] witness;
```
cprudhom: There is no need for witnesses (or watch literal) to be backtrackable.
Indeed, if a value was considered as watched at a given time, it remains true on backtrack

solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
```diff
@@ -0,0 +1,164 @@
...
+    private ISet mandatoryValues;
+    private ISet possibleValues;
+    private TIntArrayList listForRandomPick;
+    private Random rnd = new Random(0);
+
+    public PropNValue(IntVar[] vars, IntVar nvalue) {
+        super(ArrayUtils.concat(vars, nvalue), PropagatorPriority.LINEAR, true);
+        this.nValue = nvalue;
+        n = vars.length;
+        TIntArrayList list = new TIntArrayList();
```
cprudhom: Why not using a `TIntHashSet` instead ?

solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
```diff
@@ -0,0 +1,164 @@
...
+                }
+            }
+        }
+        list.sort();
+        concernedValues = list.toArray();
+        possibleValues = SetFactory.makeStoredSet(SetType.BITSET, 0, model);
+        mandatoryValues = SetFactory.makeStoredSet(SetType.BITSET, 0, model);
+        witness = new IStateInt[concernedValues.length];
+        listForRandomPick = new TIntArrayList();
+        for(int j = 0; j<witness.length; j++) {
```
cprudhom: You should call `selectRandomWitness(int idxConcernedValue)` to limit code redundancy.

solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
```diff
@@ -0,0 +1,164 @@
...
+                if(!list.contains(value)) {
+                    list.add(value);
+                }
+            }
+        }
+        list.sort();
+        concernedValues = list.toArray();
+        possibleValues = SetFactory.makeStoredSet(SetType.BITSET, 0, model);
+        mandatoryValues = SetFactory.makeStoredSet(SetType.BITSET, 0, model);
+        witness = new IStateInt[concernedValues.length];
```
cprudhom: Can be initialized with 
```java
witness = new int[concernedValues.length];
Arrays.fill(witness, -1);
```

solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
ArthurGodet: Indeed, I should use int[] instead of IStateInt[]

solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
ArthurGodet: Because there is no method sort() on a TIntHashSet, and I think there might be code optimisation to do if the values are sorted.

solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
ArthurGodet: Indeed, I forgot to change this code

solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
cprudhom: But why to you the values in `concernedValues ` to be sorted ?
Alternatively, you could sort `concernedValues` in place using `Arrays.sort`.

PR #676, opened 2020-03-18
Add ConflictHistorySearch blackbox heuristic

cprudhom:
Changes proposed in this PR:
- add new black box search heuristic, named `ConflictHistorySearch`

[Conflist History Based Branching Heuristic for CSP Solving](http://www.lsis.org/terriouxc/fr/publis/cima2018.pdf)


solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/ConflictHistorySearch.java
```diff
@@ -0,0 +1,133 @@
...
+    @Override
+    public void onContradiction(ContradictionException cex) {
+        if (cex.c instanceof Propagator) {
+            Propagator p = (Propagator) cex.c;
+            double qj = q.get(p);
+            // update q
+            q.put(p, (1 - a) * qj + a * r.get(p));
+            // update r
+            double rj = r.get(p);
+            r.put(p, 1d / (conflicts - rj + 1));
```
ArthurGodet: I do not agree with the update of the reward. According to the paper, the reward used inside the update of q(cj) uses the reward (1d / (conflicts - rj + 1)) whereas rj is simply the latest value of variable conflicts when this specific constraint (here Propagator) fails.

So, for me, the code should be : `r.put(p, 1d / (conflicts - conflict.get(p) + 1));`

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/ConflictHistorySearch.java
```diff
@@ -0,0 +1,133 @@
...
+        double w = 0.;
+        int nbp = v.getNbProps();
+        for (int i = 0; i < nbp; i++) {
+            Propagator prop = v.getPropagator(i);
+            int pid = prop.getId();
+            int futVars = pid2arity.get(pid);
+            if (futVars == -1) {
+                futVars = prop.arity();
+                pid2arity.put(pid, futVars);
+            }
```
ArthurGodet: I do not get the idea behind pid2arity. In the paper, the weight is increased by q.get(prop) + D only if its number of uninstantiated variables is strictly greater than one. Shouldn't we do the same behaviour with a method that would look like this ?

`
private static boolean hasStrictlyMoreThanOneUninstantiatedVariables(Propagator prop) {
        int nbUninstantiatedVariables = 0;
        for(int i = 0; i<prop.getNbVars(); i++) {
            if(!prop.getVar(i).isInstantiated()) {
                nbUninstantiatedVariables++;
                if(nbUninstantiatedVariables > 1) {
                    return true;
                }
            }
        }
        return false;
    }
`

Or is it too costful on the long run ?

solver/src/test/java/org/chocosolver/solver/search/CHSTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.chocosolver.solver.Model;
+import org.chocosolver.solver.Solver;
+import org.chocosolver.solver.search.limits.FailCounter;
+import org.chocosolver.solver.search.strategy.Search;
+import org.chocosolver.solver.variables.IntVar;
+import org.chocosolver.util.ProblemMaker;
+import org.chocosolver.util.tools.ArrayUtils;
+import org.testng.annotations.Test;
+
+import static org.chocosolver.solver.search.strategy.Search.conflictHistorySearch;
```
ArthurGodet: Static import should not be used (especially if Search is imported anyway)

solver/src/test/java/org/chocosolver/solver/search/DOWDTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.chocosolver.solver.Model;
+import org.chocosolver.solver.Solver;
+import org.chocosolver.solver.search.limits.FailCounter;
+import org.chocosolver.solver.search.strategy.Search;
+import org.chocosolver.solver.variables.IntVar;
+import org.chocosolver.util.ProblemMaker;
+import org.chocosolver.util.tools.ArrayUtils;
+import org.testng.annotations.Test;
+
+import static org.chocosolver.solver.search.strategy.Search.domOverWDegSearch;
```
ArthurGodet: Same remark about static imports

solver/src/test/java/org/chocosolver/solver/search/CHSTest.java
```diff
@@ -0,0 +1,61 @@
...
+//		Model s1 = costasArray(7, false);
+		Model s2 = costasArray(7, true);
+
+//		while (s1.getSolver().solve()) ;
+//		out.println(s1.getSolver().getSolutionCount());
+
+		while (s2.getSolver().solve()) ;
+
+//		out.println(s2.getSolver().getSolutionCount());
+//		assertEquals(s1.getSolver().getSolutionCount(), s2.getSolver().getSolutionCount());
```
ArthurGodet: What is the interest of the test if this line is commented ?

solver/src/test/java/org/chocosolver/solver/search/DOWDTest.java
```diff
@@ -0,0 +1,61 @@
...
+//		Model s1 = costasArray(7, false);
+		Model s2 = costasArray(7, true);
+
+//		while (s1.getSolver().solve()) ;
+//		out.println(s1.getSolver().getSolutionCount());
+
+		while (s2.getSolver().solve()) ;
+
+//		out.println(s2.getSolver().getSolutionCount());
+//		assertEquals(s1.getSolver().getSolutionCount(), s2.getSolver().getSolutionCount());
```
ArthurGodet: Same remark about interest of the test

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/ConflictHistorySearch.java
cprudhom: There is something to do here, I agree. I will suggest something else.

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/ConflictHistorySearch.java
```diff
@@ -93,8 +93,7 @@ public void onContradiction(ContradictionException cex) {
             // update q
             q.put(p, (1 - a) * qj + a * r.get(p));
             // update r
-            double rj = r.get(p);
-            r.put(p, 1d / (conflicts - rj + 1));
+            r.put(p, 1d / (conflicts - conflict.get(p) + 1));
```
ArthurGodet: I think that updating r should be done after updating conflict map

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/AbstractCriterionBasedStrategy.java
```diff
@@ -105,4 +106,22 @@ public AbsctractCriterionBasedStrategy(IntVar[] vars, long seed,
...
+        int futVars = pid2arity.get(pid);
+        if (futVars == -1) {
+            futVars = 0;
+            for (int i = 0; i < prop.getNbVars(); i++) {
+                if (!prop.getVar(i).isInstantiated()) {
+                    if (++futVars > 1) {
+                        break;
+                    }
+                }
+            }
```
ArthurGodet: What is the purpose of this for loop ? As I understand it, it increases futVars until it is strictly greater than 1 and then call arity() that is basically this for loop, without the stopping criterion.

cprudhom: Arf ! You're right, maybe I should stop working with my children around me

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/ConflictHistorySearch.java
cprudhom: Well, I don't think so: " Once r(ci) and q(ci) are updated, #Conflicts is incremented by 1."
Moreover, if we update conflict before, r will always be equal to 1, for all propagators.

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/AbstractCriterionBasedStrategy.java
cprudhom: But I claim that considering big arity constraint it is worth computing and storing that value once.

ArthurGodet: Well, I would if this value was not moving all the time. The number of uninstantiated variables inside a propagator might be different at each node depending on the filterings that are applied.

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/ConflictHistorySearch.java
ArthurGodet: Yes, I agree that variable `conflicts` should be incremented at the end. But `r` map is one step later from its real value with the update in this order. And, actually, I think that `r` map should be removed as saving the value of the reward is not useful because it is easy to compute and depends only on `conflicts` and `conflict.get(p)`

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/AbstractCriterionBasedStrategy.java
cprudhom: But `pid2arity` is cleared l.66, before iterating over variables. 
The only purpose of `pid2arity` is simply to avoid  to compute the "arity" of a propagator everytime a variable of his scope needs to know its weight.

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/ConflictHistorySearch.java
cprudhom: You're right about removing `r` which is useless in that context

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/AbstractCriterionBasedStrategy.java
ArthurGodet: Yes, my bad, it is a good idea to have this map !

PR #678, opened 2020-03-28
Correct the `keySort()` example

mrattfeldt:
The permutation argument reflects the sorted position (1 based) of the
i:th element in `vars`.

Fixes # .

Changes proposed in this PR:
- 
- 
- 
 
@chocoteam/core-developer


cprudhom:
Travis fails, but it has to do with something else (I think the right to run some specific steps in the script). So I force PR merge.

PR #685, opened 2020-04-17
Rework Maven structure

ArthurGodet:
Changes proposed in this PR:
- Rework of the pom.xml (add two new profiles + all reporting section)
- Add checkstyle to the project (to organise future improvements of code)

For our Maven organisation, this is a beginning and I think there is still room for improvement
 
@chocoteam/core-developer


pom.xml
```diff
@@ -68,6 +69,12 @@
             <organization>COSLING S.A.S.</organization>
             <organizationUrl>http://www.cosling.com</organizationUrl>
         </developer>
+        <developer>
```
cprudhom, line 72: ^^

pom.xml
```diff
@@ -138,6 +145,30 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
```
cprudhom, line 148: Can you explain what those two profiles are for ?

pom.xml
```diff
@@ -282,7 +363,168 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-site-plugin</artifactId>
+                <version>3.4</version>
+            </plugin>
         </plugins>
     </build>
 
+    <!-- =============================================================== -->
+    <!-- Gestion des rapports -->
```
cprudhom, line 379: In English please :)

pom.xml
ArthurGodet: noIbex profile is used to launch tests of the groups 1s, 10s and checker, avoiding build to fail locally when Ibex is not installed

dev profile is here to have a default profile as teh two others are declared not active by default.

pom.xml
ArthurGodet: Sorry, it was a hasty copy-paste. It is in english now

pom.xml
cprudhom: Ok. How to we use the profiles?

ArthurGodet: In command line, mvn has a parameter to select the profile: -P. So let say you want to locally build (and install in your local maven repository) Choco Solver using the noIbex profile because you don't have Ibex installed on your computer. You can do it with the following line:
`mvn clean install -P noIbex` or `mvn clean install -PnoIbex`

PR #686, opened 2020-04-17
Add time measure to first/best solution inside Solver

ArthurGodet:
Changes proposed in this PR:
- Add a time measure in Solver: this is the time to first/best solution
- Remove some dead code + Improve some tests code
 
@chocoteam/core-developer


solver/src/main/java/org/chocosolver/solver/objective/IBoundsManager.java
```diff
@@ -46,27 +46,29 @@ default boolean isOptimization() {
...
-     *
-     * @param b lower and upper bounds
-     */
-    default void updateBestBounds(IBoundsManager b) {
-        if(isOptimization()) {
-            updateBestLB(b.getBestLB());
-            updateBestUB(b.getBestUB());
-        }
-    }
+    //    /**
```
cprudhom, line 61: Why is this part commented?

ArthurGodet: It was not used anywhere in the code according to IntelliJ so I commented it to see if everything was still working locally and I forgot to delete this dead code.

cprudhom: I think this is related to [eps4j-mpj-choco](https://github.com/eps4j/eps4j-mpj-choco), but the repo is not active anymore, so let's remove it.

solver/src/main/java/org/chocosolver/solver/search/measure/IMeasures.java
```diff
@@ -147,8 +159,9 @@ default String toOneLineString() {
         if (hasObjective()) {
             st.append(getBoundsManager()).append(", ");
         }
-        st.append(String.format("Resolution time %.3fs, %d Nodes (%,.1f n/s), %d Backtracks, %d Backjumps, %d Fails, %d Restarts",
+        st.append(String.format("Resolution time %.3fs, Time to first/best solution %.3fs, %d Nodes (%,.1f n/s), %d Backtracks, %d Backjumps, %d Fails, %d Restarts",
```
cprudhom, line 162: Now, that I think about it, I wonder why, in satisfaction, the time of the first solution is recorded.
Could we simply ignore this value in sat?

ArthurGodet: I wanted to adapt the text between first and best but there is no way (with the current state of code) to access if it is a satisfaction or optimisation problem inside IMeasures.

That said, I join you in the case of satisfaction, recording this value is not useful. But we fall once again in the problem of IMeasures not able to access to the status of the problem (satisfaction or optimisation).

Any idea ?

cprudhom: `IMeasures#hasObjective()` is what you are looking for ;)
See for instance what is done in `IMeasures.toCSV()` (l.254)

PR #687, opened 2020-05-04
Fix PropNValue

ArthurGodet:
Changes proposed in this PR:
- fix PropNValue when nValue variable instantiated when mandatoryValues set is empty
 
@chocoteam/core-developer


solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
```diff
@@ -34,7 +35,7 @@
     private IntVar nValue;
     private final int n;
     private int[] concernedValues;
-    private int[] witness;
+    private IStateInt[] witness;
```
cprudhom, line 38: hmmm.. this shouldn't be backtrackable !
This is a regression

cprudhom (dismissed):

solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropNValue.java
ArthurGodet: Making witness array backtrackable gives a quicker propagator (at least for the corresponding test) and ease backtracking.

cprudhom: that's a nonsense: witness (or watch literal) are natively backtrackable.
Indeed, a witness points to a supported value, which is always true on backtrack, so no need to make backtrackable.
If that is quicker, you should think of why it is, because since you are doing less operations, it shouldn't be quicker.

ArthurGodet: You are right. It was the getting back to -1 for the witness that was posing problems. Now, it should be good, and the propagator is indeed faster with an int[] array

PR #709, opened 2020-08-23
made combination fo strategysequencer and fap compile. Can't test as …

glelouet:
…there are too many compilation error in the project.


cprudhom:
If you don't mind, I will manually merge your modifications into `master`.

glelouet:
I could not test them, I quickly looked at the commit and it seems you made a test so I guess it's working :)

Of course merge what you want, it's more to provide example than to be integrated as is.

Issue #702, opened 2020-08-11
[BUG] Arithmetic bug with Ibex and power operator


schmittjoaopedro:
I'm having a bug with the following program:

```java
Model model = new Model();
RealVar x = model.realVar("x", -10000, 10000, 0.001);
RealVar y = model.realVar("y", -10000, 10000, 0.001);

y.eq(x.pow(2.0)).ibex(0.001).post();
model.getSolver().propagate();
System.out.println(x); // Domain should be -> [-100 .. 100], but returns -> [0 .. 100]
```
## Expected behavior

The domain of X is not being filtered correctly. It should return the negative part (-100.0 .. 0) as well.

## Context
I'm using this for product configuration, it's very important because influences in the dimensional calculations of engineering products. I found this error during a calculation of physical properties between two different mechanical componentes.

## Environment
* Choco-solver version: 4.10.3
* Java version: 1.8  x32 bits

cprudhom:
Hi,

I suppose, this is related to Ibex:
```java
Ibex ibex = new Ibex(new double[]{0.001, 0.001, 0.001});
        
ibex.add_ctr("{0}={1}^{2}");
ibex.build();
double domains[] = {-16.0, 16.0, -4.0, 4.0, 2.0, 2.0};
System.out.println("Before contract:");
System.out.println("([" + domains[0] + "," + domains[1] + "] ; [" + domains[2] + "," + domains[3] + "])");
        
int result = ibex.contract(0, domains);
        
if (result == Ibex.FAIL) {
    System.out.println("Failed!");
} else if (result == Ibex.CONTRACT) {
    System.out.println("After contract:");
    System.out.println("([" + domains[0] + "," + domains[1] + "] ; [" + domains[2] + "," + domains[3] + "])");
} else {
    System.out.println("Nothing.");
}
ibex.release();
```
Let's check with them.

cprudhom:
Follow [Ibex issue](https://github.com/ibex-team/ibex-lib/issues/471)

schmittjoaopedro:
As commented by @gchabert in the issue [IBEX-LIB-471](https://github.com/ibex-team/ibex-lib/issues/471), the problem is related about how choco-solver sends the power operations to ibex, as ibex can't handle complex numbers (it only work with positive numbers in the domain of power functions), then the expressions should be changed from:
```java
model.realIbexGenericConstraint("{0}={1}^{2}", y, x, k).post();
```
to:
```java
[...]

PR #710, opened 2020-09-03
Fix arithmetic bug with Ibex and power operator for integer exponents

schmittjoaopedro:
This changes proposes a handling for integer exponents of the continuous power function (#702). 

Fixes # .

Changes proposed in this PR:
- Changed `pow` function of `BiCArExpression` to identify integer exponents and create the right expression for ibex, where the exponent is declared as a constant.
- Create a test to simulate the scenario
- Fix broken other tests
 
@chocoteam/core-developer

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/expression/continuous/arithmetic/BiCArExpression.java
+++ b/solver/src/main/java/org/chocosolver/solver/expression/continuous/arithmetic/BiCArExpression.java
@@ -283,6 +283,6 @@ public class BiCArExpression implements CArExpression {
     }
 
     private boolean isIntegerConstant(RealVar realVar) {
-        return realVar.isAConstant() && (realVar.getLB() % 1) == 0;
+        return realVar.isAConstant() && Math.rint(realVar.getLB()) == realVar.getLB();
     }
 }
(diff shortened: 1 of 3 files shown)


solver/src/main/java/org/chocosolver/solver/expression/continuous/arithmetic/BiCArExpression.java
```diff
@@ -276,4 +281,8 @@ public void intersect(double lb, double ub, ICause cause) throws ContradictionEx
     public String toString() {
         return op.name() + "(" + e1.toString() + "," + e2.toString() + ")";
     }
+
+    private boolean isIntegerConstant(RealVar realVar) {
+        return realVar.isAConstant() && (realVar.getLB() % 1) == 0;
```
cprudhom, line 286: What about using this:
`&& Math.rint(realVar.getLB()) == realVar.getLB()`
Regarding the `rint(a)`Javadoc:
> Special cases:
     If the argument value is already equal to a mathematical integer, then the result is the same as the argument.

cprudhom (changes requested):

solver/src/main/java/org/chocosolver/solver/expression/continuous/arithmetic/BiCArExpression.java
schmittjoaopedro: Sure, no problem. I've commited the change

cprudhom (approved):

Issue #509, opened 2017-03-21
QuickXplain in Choco


jkassem:
Hi,

I'm a beginner in using Choco so any help would be higly appreciated. I want to ask if QuickXplain algorithm is already implemented in Choco or not? If not, how to start with it, what to do, is there any functions that I can use that would make the implementation easier?

Thanks.

cprudhom:
No, QuickXplain is not implemented in Choco, and I never ask myself how to integrate it in choco.
I don't think I'll time to do it, but if you feel like implementing it for Choco, I'll do my best to help you.

However, keep in mind that Choco provides CBJ which may be useful for over constrained problems too.

cprudhom:
Did you start working on QuickXplain ?

jkassem:
I'm still trying to be more familiar with Choco... Since I'm new to Choco and as I see there's no comprehensive guide nor examples, am still trying in the library and checking the available classes & functions.... I checked CBJ but the same, no enough explanation (in my point of view).
If you can guide me or recommend me specific helpful things to read and try, I would be grateful.

Thank you

cprudhom:
You can find [documentation](http://choco-solver.readthedocs.io/en/latest/?badge=latest), [tutorials](http://choco-tuto.readthedocs.io/en/latest/) and [javadoc](http://www.choco-solver.org/apidocs/index.html) links on the [website of choco](http://choco-solver.org).

So if I were you, I start by:
1. getting a set of unsatisfiable problems,
2. adding a new API to `IResolutionHelper` that follows QuickXplain, as depicted in [the article](https://www.aaai.org/Papers/AAAI/2004/AAAI04-027.pdf)
 
In addition:
- ignore consistency level, at least at the beginning, simply look for a solution calling `solver.solve() `.
- consider either `solver.post(c)` and `solver.unpost(c)` to add or ignore constraints.
- do not try to make it parallel.
- keep in mind that a constraint is made of one or more propagators.

If you have specific question, ask them on this thread (no google-group) or on a private chat room on Gitter.

jkassem:
Thanks for the advices, i will work on it....

jkassem:
Hi again,
When I use model.getCstrs(), I get a lot more than the posted constraints. I guess there are some constraints that are generated inside? This causes a problem when dividing the constraints (applying QuickXplain) so how can get only the constraints I posted? 
Thanks

cprudhom:
You can't get only posted constraints.
[...]

PR #750, opened 2020-12-01
Quick xplain proposal

schmittjoaopedro:
I'm proposing this PR @cprudhom so that you will be able to better evaluate what we were talking about in gitter.

Changes proposed in this PR:
- Add a QuickXPlain algorithm to find the minimum conflicting set of constraints (ISSUE #509)
- Add support to disable constraints from execution (this feature is also required for QuickXPlain)
 
@chocoteam/core-developer

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/Solver.java
+++ b/solver/src/main/java/org/chocosolver/solver/Solver.java
@@ -628,10 +628,15 @@ public class Solver implements ISolver, IMeasures, IOutputFactory {
 
     /**
      * Return the minimum conflicting set from a conflicting set that is causing contradiction.
-     * @param conflictingSet
-     * @return minimumConflictingSet
+     *
+     * @param conflictingSet the super-set of constraints causing contradiction
+     * @return minimumConflictingSet of constraints (the root cause of contradiction)
+     * @throws SolverException when MCS is called during solving
      */
     public List<Constraint> findMinimumConflictingSet(List<Constraint> conflictingSet) {
+        if (isSolving()) {
+            throw new SolverException("Minimum Conflicting Set (MCS) can't be executed during solving");
+        }
         return new QuickXPlain(getModel()).findMinimumConflictingSet(conflictingSet);
     }
 
@@ -800,6 +805,20 @@ public class Solver implements ISolver, IMeasures, IOutputFactory {
             head = dpath.getLastDecision();
         }
         return success;
+    }
+
+    /**
+     * Solving is executing if the search state is different from NEW, that is,
+     * if it has started to branch decisions.
+     * A double check for execution is done looking if the environment trailing
+     * has started as well.
+     *
+     * @return isSolving if the solver is executing searching or branching
+     */
+    public boolean isSolving() {
+        boolean isSearching = getSearchState() != SearchState.NEW;
+        boolean isTrailing = getEnvironment().getWorldIndex() > 0;
+        return isSearching || isTrailing;
     }
 
     ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/solver/src/main/java/org/chocosolver/solver/Solver.java
+++ b/solver/src/main/java/org/chocosolver/solver/Solver.java
@@ -627,6 +627,15 @@ public class Solver implements ISolver, IMeasures, IOutputFactory {
     }
 
     /**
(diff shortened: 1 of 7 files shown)


solver/src/main/java/org/chocosolver/solver/Solver.java
```diff
@@ -626,6 +626,15 @@ public void propagate() throws ContradictionException {
         }
     }
 
+    /**
+     * Return the minimum conflicting set from a conflicting set that is causing contradiction.
+     * @param conflictingSet
+     * @return minimumConflictingSet
+     */
+    public List<Constraint> findMinimumConflictingSet(List<Constraint> conflictingSet) {
```
cprudhom, line 634: I think this is a good place to add assertions and eventually throw exception if looking for the MCS is called in inappropriate way (during solving for instance). The javadoc should also exhibit these assertions.

solver/src/main/java/org/chocosolver/solver/constraints/Constraint.java
```diff
@@ -84,6 +86,11 @@
      */
     private String name;
 
+    /**
+     * If a constraint is enabled to the propagation engine.
+     */
+    private boolean enabled = Boolean.TRUE;
```
cprudhom, line 92: Shouldn't it be `true`? What is the benefit of using Boolean.TRUE?

solver/src/main/java/org/chocosolver/solver/constraints/Constraint.java
```diff
@@ -391,4 +398,37 @@ public static Constraint merge(String name, Constraint... toMerge) {
         }
         return new Constraint(name, props.toArray(new Propagator[0]));
     }
+
+    public boolean isEnabled() {
```
cprudhom, line 402: Could you add some javadoc?

solver/src/main/java/org/chocosolver/solver/constraints/Propagator.java
```diff
@@ -100,6 +100,11 @@
      */
     private static final short PASSIVE = 3;
 
+    /**
+     * Ignore propagation during execution.
+     */
+    private boolean enabled = Boolean.TRUE;
```
cprudhom, line 106: Same question as previous: I would have expected `true`

cprudhom (changes requested):

solver/src/main/java/org/chocosolver/solver/Solver.java
schmittjoaopedro: Sure, I add the check for solving and created a unitary test to thist.

solver/src/main/java/org/chocosolver/solver/constraints/Constraint.java
schmittjoaopedro: Sorry, it's a costume that I have.
I checked on the internet and to use the primite `true` only requires a shift in registers (that's more efficient than initialize a static class). In fixed to use primitive `true`.

solver/src/main/java/org/chocosolver/solver/constraints/Constraint.java
schmittjoaopedro: Yes, I added on the last commit.

solver/src/main/java/org/chocosolver/solver/constraints/Propagator.java
schmittjoaopedro: Fixed.

cprudhom (approved):

solver/src/main/java/org/chocosolver/solver/Solver.java
```diff
@@ -793,6 +807,20 @@ public boolean moveBackward(){
...
+     * Solving is executing if the search state is different from NEW, that is,
+     * if it has started to branch decisions.
+     * A double check for execution is done looking if the environment trailing
+     * has started as well.
+     *
+     * @return isSolving if the solver is executing searching or branching
+     */
+    public boolean isSolving() {
+        boolean isSearching = getSearchState() != SearchState.NEW;
+        boolean isTrailing = getEnvironment().getWorldIndex() > 0;
```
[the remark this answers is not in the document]
schmittjoaopedro: Fixed.

schmittjoaopedro:
I've taken the freedom to add the QuickXPlain feature item in `CHANGES.md`

cprudhom (approved):

PR #765, opened 2021-02-22
Record only some variables in findAllSolutions

MathieuVavrille:
Changes proposed in this PR:
- Adding function to find all solutions, but store only some variables of the problem in the solutions
 
@chocoteam/core-developer


cprudhom:
As I told you orally, I think it would be a better idea to store this kind of information directly in the Model or in variables instead of changing those API. Plus, some files are comited even they are not part of the PR (ie, FlatZinc files)

PR #768, opened 2021-03-05
Migrate graph variables from choco-graph - Refactor graph API - Implement set views over graph variables

dimitri-justeau:
Fixes # .

Changes proposed in this PR:
- Graph variables were migrated from choco-graph to choco, with some graph API refactoring. This refactoring does not change the behaviour of graph data structures nor graph variable, but attempt to provide a cleaning and unambiguous API for using graph data structures and variable. The changes are the following:

    - The semantic distinction between arcs and edges has been removed for more clarity (arc is mainly used in French, but in     English it is much more common to talk about (directed or not) edges. If the graph is undirected, edges are undirected, if the graph is directed, the graph is directed.
        - In graph data structures (`IGraph`), methods related to edges are `addEdge`, `removeEdge`, and `containsEdge`.
        - In graph variables (`GraphVar`), methods related to edges are `enforceEdge` and `removeEdge`.
 
    - The object model is such that the more abstract interface specifies directed graph accessors on edges (in `IGraph`: `getSuccessorsOf` and `getPredecessorsOf`; in GraphVar: `getMandatorySuccessorsOf`, `getPotentialSuccessorsOf`, `getMandatoryPredecessorsOf`, and `getPotentialPredecessorsOf`. When the graph is undirected, the methods `getNeighbors` in `UndirectedGraph`, and `getMandatoryNeighborsOf` and `getPotentialNeighborsOf` in `UndirectedGraphVar` are available. Note that an undirected graph is equivalent to a directed graph with couples of opposite directed edges. Thus, the neighbors of a node in an undirected graph are both successors and predecessors, and this is why these methods are equivalent to neighbors related methods in the case of an undirected graph. To encourage unambiguous use and facilitate code reading, the successors and predecessors related method have been defined as deprecated in explicit uses of UndirectedGraphs.
 
   - The possibility to chose (in graph data structure constructors) and get the set data structure for nodes has also been added, as it was only implemented for neighbors: `getNodeSetType` and `getEdgeSetType`. The previous default behaviour has been conserved with default constructors. 
- Set views over graph variables were implemented and tested. It is planned to add more of these views as they provide straightforward and expressive modelling possibilities. For now, the abstract class `GraphSetView` and the views `GraphNodeSetView` (set view over the nodes of a graph var), `GraphSuccessorsSetView` (set view over successors of a node of a graph, or simply neighbors for undirected graph vars), and `GraphPredecessorsSetView` (same as previous with predecessors) were implemented.
- `GraphDeltaMonitor` has been modified in accordance with commit 84a3fc304ceca684cebe776543a3bb4e5b6d76ad.
- All new functionalities over graph variables and graph views were integrated in corresponding factoring following the coding style of Choco.
- A `GraphFactory` was implemented, similarly to the `SetFactory`, to provide useful functions for instantiating graph data structures.
 
@chocoteam/core-developer

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/search/strategy/Search.java
+++ b/solver/src/main/java/org/chocosolver/solver/search/strategy/Search.java
@@ -156,7 +156,7 @@ public class Search {
 
     /**
      * Lexicographic graph branching strategy.
-     * Branch on nodes then arcs/edges.
+     * Branch on nodes then edges.
      * <br>
      * <br> node branching:
      * Let i be the first node such that
@@ -164,9 +164,9 @@ public class Search {
      * The decision adds i to the kernel of g.
      * It is fails, then i is removed from the envelope of g.
      * <br>
-     * arc/edge branching:
+     * edge branching:
      * <br> node branching:
-     * Let (i,j) be the first arc/edge such that
+     * Let (i,j) be the first edge such that
      * (i,j) in envelope(g) and (i,j) not in kernel(g).
      * The decision adds (i,j) to the kernel of g.
      * It is fails, then (i,j) is removed from the envelope of g
--- a/solver/src/main/java/org/chocosolver/solver/search/strategy/Search.java
+++ b/solver/src/main/java/org/chocosolver/solver/search/strategy/Search.java
@@ -26,10 +26,7 @@ import org.chocosolver.solver.search.strategy.decision.IbexDecision;
 import org.chocosolver.solver.search.strategy.selectors.values.*;
 import org.chocosolver.solver.search.strategy.selectors.variables.*;
 import org.chocosolver.solver.search.strategy.strategy.*;
-import org.chocosolver.solver.variables.IntVar;
-import org.chocosolver.solver.variables.RealVar;
-import org.chocosolver.solver.variables.SetVar;
-import org.chocosolver.solver.variables.Variable;
+import org.chocosolver.solver.variables.*;
 import org.chocosolver.util.bandit.MOSS;
 import org.chocosolver.util.bandit.Static;
 import org.chocosolver.util.tools.VariableUtils;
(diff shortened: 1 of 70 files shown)


jgFages (approved):
Good initiative!
Maybe the documentation should be updated as well

solver/src/main/java/org/chocosolver/solver/search/strategy/Search.java
```diff
@@ -153,6 +150,33 @@ public static SetStrategy setVarSearch(SetVar... sets) {
...
+     * edge branching:
+     * <br> node branching:
+     * Let (i,j) be the first edge such that
+     * (i,j) in envelope(g) and (i,j) not in kernel(g).
+     * The decision adds (i,j) to the kernel of g.
+     * It is fails, then (i,j) is removed from the envelope of g
+     *
+     * @param g a graph variable to branch on
+     */
+    public static GraphStrategy graphVarLexSearch(GraphVar g) {
```
cprudhom, line 176: We could consider, in another PR, to change this name to a shorter one

solver/src/main/java/org/chocosolver/solver/variables/view/IView.java
```diff
@@ -38,7 +37,7 @@
      *
      * @return variable observed
      */
-    IntVar getVariable();
+    Variable getVariable();
```
cprudhom, line 40: We can consider generifying the class, don't you think ? Well, it can be declared in a distinct issue, since that's not central to this PR

cprudhom (approved):

dimitri-justeau:
> Good initiative!
> Maybe the documentation should be updated as well

Thank you Jean-Guillaume ! The plan is to migrate all features of Choco-graph into Choco, with some improvements, notably the views on graph variables that will provide the same functionalities as channelling constraints, with memory and performances gains.

solver/src/main/java/org/chocosolver/solver/search/strategy/Search.java
dimitri-justeau: Yes, would you like to do it now, or should we wait until the moment in which we will focus on graph strategies?

solver/src/main/java/org/chocosolver/solver/variables/IVariableFactory.java
```diff
@@ -819,6 +821,37 @@ default SetVar setVar(String name, int... value) {
...
+     * Creates an undirected graph variable, taking its values in the graph domain [LB, UB].
+     * An instantiation of a graph variable is a graph composed of nodes and edges.
+     * LB is the kernel graph (or lower bound), that must be a subgraph of any instantiation.
+     * UB is the envelope graph (or upper bound), such that any instantiation is a subgraph of it.
+     * @param name Name of the variable
+     * @param LB The lower bound graph (or kernel)
+     * @param UB The upper bound graph (or envelope)
+     * @return An undirected graph variable taking its values in the graph domain [LB, UB].
+     */
+    default UndirectedGraphVar undirectedGraphVar(String name, UndirectedGraph LB, UndirectedGraph UB) {
```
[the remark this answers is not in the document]
dimitri-justeau: I proposed these names to follow the object model, which defines directed graph methods at the more abstract level, but these names are indeed shorter and probably more common in graph APIs, so I'm ok to change the names in another PR.

solver/src/main/java/org/chocosolver/solver/variables/view/IView.java
dimitri-justeau: I totally agree, I will do a small PR for that.

solver/src/main/java/org/chocosolver/solver/search/strategy/Search.java
cprudhom: I can wait until you work on strategies

PR #789, opened 2021-04-20
Implement and test Graph variable induced from the envelope.

dimitri-justeau:
Fixes # .

Changes proposed in this PR:
- Implement and test a specific class of graph variable that is induced from the envelope that is used to construct it. Decisions can only occur on nodes, as two nodes included in the graph variable are connected if and only if they are connected in the graph that was used as the envelope to instantiate the variable. Such graphs are useful when this property is required by the problem specifications, as they allow a drastic decrease in combinatorial complexity. See `nodeInducedGraphVar` method in `IVariableFactory`.
 
@chocoteam/core-developer


cprudhom:
It seems that this PR is an extension of #788 ?

dimitri-justeau:
> It seems that this PR is an extension of #788 ?

More or less yes. The feature introduced in this PR is not a view but a more restrictive implementation of GraphVar that can provide performances improvements when its specifications correspond to the formulation of the problem (the graph must be an induced-subgraph of the initial envelope). This is different from subgraph views as there is no notion of subset of nodes to exclude or intersect, it is based on the envelope used to instantiate the variable.

cprudhom (approved):

cprudhom (approved):

PR #790, opened 2021-04-20
Implement GraphEvtScheduler and make GraphStrategy consistent with Choco's design

dimitri-justeau:
Fixes # .

Changes proposed in this PR:
- Implement GraphEvtScheduler to ensure efficient calls to graph variables' propagators according with GraphEvents.
- Modify the design of GraphStrategy to make it consistent with Choco's design for other variables types. A GraphStrategy is constructed using the following generic constructor:
```    /**
     * Generic strategy to branch on graph variables
     *
     * @param varS          Variable selection strategy
     * @param nodeOrEdgeS   Node or edge selection (defines if whenever a decision must be on nodes or edges)
     * @param nodeS         Node selector (defines which node to enforce/remove if decision is on nodes)
     * @param edgeS         Edge selector (defines which edge to enforce/remove if decision is on edges)
     * @param enforceFirst  branching order true = enforce first; false = remove first
     * @param graphs        GraphVar array to branch on
     * @return
     */
    public static GraphStrategy graphVarSearch(VariableSelector<GraphVar> varS, GraphNodeOrEdgeSelector nodeOrEdgeS,
                                               GraphNodeSelector nodeS, GraphEdgeSelector edgeS, boolean enforceFirst,
                                               GraphVar... graphs) {
        return new GraphStrategy(graphs, varS, nodeOrEdgeS, nodeS, edgeS, enforceFirst);
    }
```
The `GraphNodeOrEdgeSelector` role is to determine whether a decision must be on a node or on an edge, the `GraphNodeSelector` role is to select the node when a decision must be made on nodes, and the `GraphEdgeSelector` has the same role for edges. Existing graph strategies were adapted to this design.

 
@chocoteam/core-developer


cprudhom:
Here again, this PR depends on #788 and #789, am I right?

dimitri-justeau:
> Here again, this PR depends on #788 and #789, am I right?

Exactly ! I rebased every PR according to the updates in #788

solver/src/main/java/org/chocosolver/solver/variables/view/graph/directed/DirectedEdgeInducedSubgraphView.java
```diff
@@ -103,7 +103,7 @@ protected boolean doRemoveNode(int node) throws ContradictionException {
         if (enforceNodes.contains(node)) {
             contradiction(this, "Try to remove mandatory node");
         }
-        return false;
+        return true;
```
cprudhom, line 106: Hmmm, returning `true` suppose that a modification occurs, which is not the case here, right?

solver/src/main/java/org/chocosolver/solver/variables/view/graph/undirected/EdgeInducedSubgraphView.java
```diff
@@ -101,7 +101,7 @@ protected boolean doRemoveNode(int node) throws ContradictionException {
         if (enforceNodes.contains(node)) {
             contradiction(this, "Try to remove mandatory node");
         }
-        return false;
+        return true;
```
cprudhom, line 104: Same comment as previous one.

cprudhom (approved):

solver/src/main/java/org/chocosolver/solver/variables/view/graph/directed/DirectedEdgeInducedSubgraphView.java
dimitri-justeau: You are right, it is a non-intuitive shortcut. Actually, to remove a node from an edge-induced subgraph view, it is only necessary to isolate the node by removing every of its incident edge, so all the work is done in the `removeNode` method of the `UndirectedGraphView` (resp. `DirectedGraphView`). As you can see in the following code, the `doRemoveNode` method will only be called if there is no mandatory incident edge (`getMandatoryNodes().contains(node)`) and if there is at least one potential incident edge (`getPotentialNodes().contains(node)`). Thus, when `doRemoveNode(node)` occurs, all incident edges should have already been removed, and the node is consequently also removed, so `doRemoveNode` will always return true. I pushed a new commit with a semantically correct return statement: `return !getPotentialNodes().contains(node);`.

```
    @Override
    public boolean removeNode(int node, ICause cause) throws ContradictionException {
        assert cause != null;
        assert (node >= 0 && node < getNbMaxNodes());
        if (getMandatoryNodes().contains(node)) {
            this.contradiction(cause, "remove mandatory node");
            return false;
        } else if (!getPotentialNodes().contains(node)) {
            return false;
        }
        ISet nei = getPotentialNeighborsOf(node);
        for (int i : nei) {
            removeEdge(node, i, cause);
        }
        if (doRemoveNode(node)) {
            if (reactOnModification) {
                delta.add(node, GraphDelta.NODE_REMOVED, cause);
            }
            GraphEventType e = GraphEventType.REMOVE_NODE;
            notifyPropagators(e, cause);
            return true;
        }
        return false;
    }
```

solver/src/main/java/org/chocosolver/solver/variables/view/graph/undirected/EdgeInducedSubgraphView.java
dimitri-justeau: See previous answer.

cprudhom (approved):

PR #801, opened 2021-05-01
fix connected et nbCC

jgFages:
Fix articulation points and bridge linking mandatory nodes detection
Thanks @dimitri-justeau for the test suite!


dimitri-justeau:
> Fix articulation points and bridge linking mandatory nodes detection
> Thanks @dimitri-justeau for the test suite!

You're welcome :-) It is really cool and to have choco-graph directly into choco and I had fun doing it. Thank you for this fix, the connectivity constraints are really useful for my research and I am sure it is also the case for many users.

dimitri-justeau (approved):

PR #809, opened 2021-05-28
Implement and test Graph union view - Fix SetUnionView and SetDifferenceView - Fix Undirected/DirectedNodeInducedGraphVarImpl

dimitri-justeau:
- Implement and test graph union view, a graph view representing the union (nodes and edges) of an arbitrary set of graphs.
- Fix SetUnionView and SetDifferefnceView: the `doForceSetElement` in `SetUnionView` and `doRemoveSetElement` in `SetIntersectionView` can have multiple support (e.g. in union, a value can be remove from observed variables iff there is only one variable with this value in envelope). When this case occurs, the view must store the values to add (or remove) in order to effectively remove it when the necessary addition (or removal) condition is satisfied.


dimitri-justeau:
UPDATE: also fixes a bug in Undirected/DirectedNodeInducedGraphVarImpl. The removeEdge method was incorrect and could cause constraints to filter too much. In previous behaviour `removeEdge(x, y)` was removing nodes x and y along with edge (x,y). Now, it:
    - removes edge (x, y) and node x iff x is in the envelope and y in the kernel;
    - removes edge (x, y) and node y iff y is in the envelope and x in the kernel;
    - removes only edge (x, y) if x and y are in the envelope but not in the kernel. For this last case, if `enforceNode(x)` is called, y will be removed because edge (x, y) is in the original envelope but not in the current envelope, and vice versa.

dimitri-justeau:
UPDATE: Still in Undirected/DirectedNodeInducedGraphVarImpl: set the cause of operations that are caused by the graph class and would not happen in the general case to `this`, as propagators rely on general-case properties. E.g. and edge removal can cause node removal, which is not the case in the general case. If some bounds in a propagator (e.g. `nbConnectedComponents`) are computed from the general case, the node removal event specific to the graph class can only be taken into account by the propagator if the cause of this event is not the propagator itself.

cprudhom (approved):

PR #827, opened 2021-09-22
Fix bug in directed graph

dimitri-justeau:
Fix the bug identified by @cprudhom in DirectedGraph, related to the addition/removal of an edge when Choco is executed without assertions.


cprudhom:
If that's OK for you, I will simply add the modification to DirectedGraph in my branch and push it to master (otherwise, I will have 11 unexpected commits)

dimitri-justeau:
Totally ok ;-)

cprudhom:
Fix by 2ac3e64

PR #868, opened 2022-01-17
Add table sampling resolution policy

cprudhom:
@mathieuvavrille can you have a look to this PR and tell me if you're OK with the way I integrated your work?


MathieuVavrille:
This seems to be a good implementation of the tableSampling algorithm.
In file TuplesFactory : concernet -> concerned.
In the tableSampling function :
 - I would precise that the created stream of solutions is infinite, as solutions found are not removed. For example line 719 -> `@return an infinite stream of randomly sampled solutions.` (maybe even precise to use `.limit()`, otherwise a `.collect()` operation would be an infinite loop).
 - line 763 : the list `selectedVariables` may contain duplicates of the same variable. It is not a big issue, as it may rarely happen, but does not make sense to have twice the same variable in a table. It can be solved by checking if the newly selected variable is already in the `selectedVariables` list.
 - lines 775-777 : I think that `currentConstraint` is not unposted from the model, as it was not yet added in the `added` list.

cprudhom:
@MathieuVavrille thank you.
`selectedVariables` is a `HashSet` so it prevents multiple occurrences of a single variable.

MathieuVavrille (approved):
This is perfect

PR #872, opened 2022-02-01
Adapt conditional all different constraint to allow different consistency levels

dimitri-justeau:
In the current API, the conditional all different constraint uses Regin AC algorithm and does not provide the option to change this. I adapted the constraint according to the regular all different constraint, to allow the choice of the filtering algorithm. I also set the default filtering algorithms to be the same as the regular all different constraint.

PS: The default behaviour with Regin AC algorithm was very slow in a problem I am solving, using the same default filtering scheme as the regular AllDiff resulted in a huge speed up (~8 seconds VS ~11 minutes) in my case.

PPS: Just for information, I noticed that in Regin AC algorithm implementation, the directed graph data structure uses a bitset implementation. In my experience, this implementation is slow for large sets. Bipartite and range set implementations are better at scale.


solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/conditions/PropCondAllDiffBase.java
```diff
@@ -0,0 +1,49 @@
...
+        super(variables, priority, false);
+        this.condition = condition;
+    }
+
+    /**
+     * @return The variables respecting the condition.
+     */
+    protected IntVar[] filterVariables() {
+        int nb = 0;
+        for (IntVar v : vars) {
```
cprudhom, line 35: Why the condition is checked twice?
Wouldn't be simpler to use a stream of variables and filter them on the basis of the condition?

dimitri-justeau: I just repeated what is currently done Choco: https://github.com/chocoteam/choco-solver/blob/531ab5dde8e86348181e72af46bc14bd334906db/solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/conditions/PropCondAllDiff_AC.java#L53

But I agree that a stream + a filter would be better, I fix this right now.

cprudhom (approved):

PR #889, opened 2022-03-23
Change the way propagators are stored in variables to ease swapping

cprudhom:
Because propagators are stored in an array in each variable, swapping some of them (when they become entailed) can be improved.
This PR change the way propagators are stored in variables.
There is now one array per type of reacting events, and each array is able to manage dynamic addition and dynamic deletion. Doing so, the way a propagator becomes activated on backtrack is managed by the variable directly and not the propagator itself. 
Unless otherwise stated, any propagator able to be entailed is now automatically swap (this was not the case in older versions). This can be changed, per propagator, during the call to the super constructor.
Since this impacts iterations over propagator to be scheduled, it also impacts the behavior of black-box strategies.


solver/src/main/java/org/chocosolver/solver/Settings.java
```diff
@@ -63,7 +63,7 @@
 
     private boolean enableSAT = false;
 
-    private boolean swapOnPassivate = false;
+    private boolean swapOnPassivate = true;
```
jgFages, line 66: In which case should we put false? Is there any benefit of not using it?

solver/src/main/java/org/chocosolver/solver/variables/Variable.java
```diff
@@ -133,18 +136,26 @@
      * Return the array of propagators this
      *
      * @return the array of propagators of this
+     * @deprecated see instead
```
jgFages, line 139: see streamPropagators() instead ?

solver/src/main/java/org/chocosolver/solver/constraints/nary/sat/PropSat.java
```diff
@@ -59,7 +59,7 @@ public PropSat(Model model) {
         // this propagator initially has no variable
         // adds solver.ONE to fit to the super constructor
         super(new Variable[]{model.getNbVars() > 0 ? model.getVar(0) : model.boolVar(false)},
-                PropagatorPriority.VERY_SLOW, true);
+                PropagatorPriority.VERY_SLOW, true, false);
```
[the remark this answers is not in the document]
cprudhom: That's a mistake, it should be set to 'true' by default.

cprudhom:
It appears that there is a complex bug in this version.
I can artificially produces it with a constraint with a specific behavior (moving variables in its scope) but there is no easy patch (or at least I couldn't find it yet)
I'll leave it on standby.

cprudhom:
I think I found the origin of the bug. 
This is related to the order in which backtrackable objects are popped.
Indeed, in `EnvironmentTrailing` (default environment), trails are popped in the reverse order of their creation (remark: the order may be fixed). And operations are always popped before integers. 

### Expected behavior is:

#### At world 0

|A|B|C|D|
|-|-|-|-|

s = 0, where s is an `IStateInt`

#### At world _i_
- postTemp(E)  (ie, save `()->remove(E)` in the environment).

|A|B|C|D|E|
|-|-|-|-|-|

s = 0


- E is entailed

|*E*|B|C|D|A|
|-|-|-|-|-|

s = 1 (all elements < 1 are ignored from iteration, since they are known to be inactive)

#### Backtrack to world _i-1_
1. decrement s
2. remove E from the array
Results in:

|A|B|C|D|
|-|-|-|-|

s=0

### Since `OperationTrail` is popped before `IntTrail`
#### Backtrack to world _i-1_
1. remove E from the array
2. decrement s
Results in:

|null|B|C|D|A|
|-|-|-|-|-|

s=0

The first element in the array must be skipped, which is not desirable.

I think this should raise a more general discussion about what is considered as acceptable in term of addition/deletion of propagators. I will open a dedicated discussion thread for that.

PR #891, opened 2022-03-29
Implement `findNsolutions` and `findNOptimalSolution`

dimitri-justeau:
I know that using the `streamSolutions` and `streamOptimalSolutions` can be used to limit the number of solutions, however, this might not be intuitive for users (at least it was not for me). I thus suggest two utility functions in the IResolutionHelper, to request exactly N solutions or N optimal solutions. I did not rely on the stream solution methods to implement these functions, in order to avoid the potential overhead of using streams (which will probably be minimal in most cases but could become significant if the user requests thousands or millions of solutions).

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/search/IResolutionHelper.java
+++ b/solver/src/main/java/org/chocosolver/solver/search/IResolutionHelper.java
@@ -157,6 +157,61 @@ public interface IResolutionHelper extends ISelf<Solver> {
     }
 
     /**
+     * Attempts to find N solutions of the declared satisfaction problem.
+     * <ul>
+     * <li>If the method returns an empty list: </li>
+     * <ul>
+     * <li>
+     * either a stop criterion (e.g., a time limit) stops the search before any solution has been found,
+     * </li>
+     * <li>
+     * or no solution exists for the problem (i.e., over-constrained).
+     * </li>
+     * </ul>
+     * <li>if the method returns a list with at least one element in it:</li>
+     * <ul>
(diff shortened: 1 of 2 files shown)


cprudhom:
I have mixed feelings.
There are two existing ways to do this (see [the documentation](https://choco-solver.org/docs/solving/limits/):
```java
//...
Solver s = m.getSolver();
s.limitSolutions(4);
s.findAllSolutions();
// or while(s.solve());
```
Or:
```java
Solver s = m.getSolver();
s.limitSolutions(4);
s.findAllSolutions(new SolutionCounter(model, 4));
```
I suppose that this is low documented but these should be the ways to do what you want to do.

On the other hands, I suppose that if you were not able to find it easily, there is space for improvement.
I am considering make easier the search strategy declaration, may be this could be part of the API?

dimitri-justeau:
Ok, indeed I did not know this solution. There is no need to reimplement something, which makes this PR useless. In my opinion, a shortcut method would help, but if you never had any complaints from other users it may not be necessary. What is the behaviour with `findAllOptimalSolutions`? Does the solution count limit only apply to the second part of the search (when the optimal value for the objective have been found), or also to the first part (finding the optimal value) ?

Also, when using a time counter in `findOptimalSolution` or `findAllOptimalSolutions`, the stop criterion is not applied to the overall search. For instance if the time limit is 30s, if it takes 29s to reach the first optimal solution, the enumeration part of the search also have 30s, which can end up with a total runtime of 60s.

cprudhom:
Well, you raise another issue (a bug actually).

PR #892, opened 2022-03-30
Update Set_FixedArray.java

dimitri-justeau:
Override the toArray method of fixed array -- no need to reconstruct the whole array.


solver/src/main/java/org/chocosolver/util/objects/setDataStructures/constant/Set_FixedArray.java
```diff
@@ -139,4 +139,9 @@ public int nextInt() {
 			}
 		};
 	}
+	
+        @Override
+        public int[] toArray() {
+            return values;
```
cprudhom, line 145: I think a call to `clone()` would be safer. 
Otherwise, there is a chance that one change the values in `values` ignoring the side effect.
Example:

```java
System.out.printf("%s%n", cst);
int[] values = cst.toArray();
values[0] = 3;
System.out.printf("%s%n", cst);
```
prints:
```
{-1, 1}
{3, 1}
```

Even if that's an extreme usage, we should prevent it.

cprudhom (changes requested):
The method should return a clone of the array instead of the array itself.

dimitri-justeau:
> The method should return a clone of the array instead of the array itself.

You are right! Although that's extreme usage it can lead to headaches in debugging!

PR #894, opened 2022-04-05
`IntVar` handles with `long` values...

cprudhom:
...thanks to new domain modification methods.

It is very simple approach, that makes possible to either modify already existant propagators (probably at the expense of efficiency, but this needs to be tested) or the develop new propagators.

The methods deals with`long` in parameter by checking whether or not it can be cast into an `int`.
Depending on the method,  the consequence is different.

Consequently, it allows to better deal with underflow/overflow (but not to fix it in a general way).
In particular, it is  helpful when dealing with `sum` or `times` constraints.

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
+++ b/solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
@@ -27,8 +27,6 @@ import org.chocosolver.solver.constraints.nary.PropKnapsack;
 import org.chocosolver.solver.constraints.nary.alldifferent.AllDifferent;
 import org.chocosolver.solver.constraints.nary.alldifferent.conditions.CondAllDifferent;
 import org.chocosolver.solver.constraints.nary.alldifferent.conditions.Condition;
-import org.chocosolver.solver.constraints.nary.alldifferent.conditions.PropCondAllDiffInst;
-import org.chocosolver.solver.constraints.nary.alldifferent.conditions.PropCondAllDiffAC;
 import org.chocosolver.solver.constraints.nary.alldifferentprec.PropAllDiffPrec;
 import org.chocosolver.solver.constraints.nary.among.PropAmongGAC;
 import org.chocosolver.solver.constraints.nary.automata.CostRegular;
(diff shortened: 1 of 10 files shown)


dimitri-justeau (approved):
I think this is a nice start toward handling long values for IntVars, with no impact on existing propagators. Although the perfect solution would be using a variable to handle the sum of IntVars even when it is a long, this solution already unlock some possibilities.

For instance, this will be usefull for this project : https://github.com/dimitri-justeau/flsgen

If you think it is of interest for Choco, in this project I have adapted Choco's PropSum constraint to a PropSumOfSuare constraint with long values: https://github.com/dimitri-justeau/flsgen/blob/master/src/main/java/org/flsgen/solver/choco/PropSumOfSquares.java

cprudhom:
The best solution would be to define a `LongVar` class.
Like `BoolVar` is a specialization of `IntVar` with restrictive bounds, `IntVar` would be a specialization of `LongVar`. 
`LongVar` <-- `IntVar` <-- `BoolVar`.

A `LongVar` would only define its domain with two longs, denoting the bounds. 
`LongVar` could provide the following methods:
- `boolean instantiateTo(long value, ICause cause)`
- `boolean updateLowerBound(long value, ICause cause)`
- `boolean updateUpperBound(long value, ICause cause)`
- `boolean updateBounds(int lb, int ub, ICause cause)`
- `long getValue()`
- `long getLB()`
- `long getUB()`
- `long getRange()`
- and maybe some iterators (but that would not be really useful).

I don't think it would be relevant to allow value removal, since the main idea would simply to be allow very large domain.
Note that I tried this some time ago, and it requires to change lots of code everywhere. 

Alternatively, we could consider (but that also requires to deeply revamp the code) to distinct a variable and its domain. 
I suppose other alternatives should be considered too (like: do nothing :)) and discussed either as a [project](https://github.com/chocoteam/choco-solver/projects?type=beta) or as a [discussion](https://github.com/chocoteam/choco-solver/discussions).

dimitri-justeau:
Huumm... Generics over primitives are really missing here !! Yes, a project or a discussion is probably a good starting point for this, as it seems like support for long values implies design choices that must be wisely made. I think that the retained solution should not be problematic to migrate with generics over primitive if they become available in the future. I have done some research, and I believe that specialized Java code generation with a template processor (e.g. see https://github.com/TimeAndSpaceIO/java-primitive-specializations-generator, which has been used for https://github.com/leventov/Koloboke) may be an interesting perspective to overcome the lack of generics over primitives.

cprudhom:
Maybe this PR can be integrated, as a first step toward a better solution.
Indeed, it may be useful and does not require too much modification.
What do you think?

PR #908, opened 2022-05-26
fix update bounds

jgFages:
It was possible to obtain LB > UB without contradiction using IntervalIntVarImpl.updateBounds


cprudhom:
Thank you for seeing the problem and proposing a solution. In order to have a more homogeneous code, I propose a solution which is not based on a new exception but on the sequence of actions.

jgFages:
Sure! It looks better !

PR #935, opened 2022-09-12
SparseBitSet for a backtrackable bitset with mostly 0s.

fhermeni:
This implementation for IStateBitSet targets backtrackable bitsets that mostly contains 0. For non-dense bitsets, this implementation is more efficient in terms of memory usage as bitsets are allocated per region instead of relying on a single, possibly bit array. In this context, the iteration is also faster.

### DIFF
--- a/solver/src/main/java/org/chocosolver/memory/structure/SparseBitSet.java
+++ b/solver/src/main/java/org/chocosolver/memory/structure/SparseBitSet.java
@@ -1,0 +1,489 @@
+package org.chocosolver.memory.structure;
+
+import org.chocosolver.memory.IEnvironment;
+import org.chocosolver.memory.IStateBitSet;
+
+import java.util.Arrays;
+
+/**
+ * A backtrackable bitset optimised for compaction. This implementation saves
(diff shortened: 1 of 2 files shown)


fhermeni:
This patch does not force any integration inside choco Environment given the criticality of that part. For the moment, this is just something I will have in btrplace and that may be useful for everyone.

cprudhom:
The following code throws an error:
```java
    IEnvironment env = new EnvironmentTrailing();
    IStateBitSet bs = env.makeSparseBitset(64);
    bs.set(100);
    bs.set(200);
    bs.clear(101, 199);
```
Indeed, in `SparseBitSet#clear(int,int)`, [line 286](https://github.com/chocoteam/choco-solver/blob/f1f864ef1cc1ef205f716f67cb3b09b3e922ab5c/solver/src/main/java/org/chocosolver/memory/structure/SparseBitSet.java#L286):
```java
      final int st;
      if (bIdx == blockIndex(from)) {
        st = from;
      } else {
        st = 0;
      }
```

`st = from` should be replaced by `st = localIndex(from);`

right?

PR #950, opened 2022-10-14
Dev restart

cprudhom:
This branch is dedicated to a small revamp of the way restarts are managed.
The main idea is to avoid using a `Move` implementation to deal with restarts but to integrate this directly in the `Solver` object. Doing so, it becomes easier (well, doable) to remove or replace a restart policy.
In addition, strategies and variable selectors are instrumented to indicate if they are *dynamic*, that is learn-based objects that are to be combined with restarts to be more efficient. A warning is printed to the console (if warnings are enabled) to indicate poor combinations (such as dyn. strat. without restarts).


solver/src/main/java/org/chocosolver/solver/Solver.java
```diff
@@ -445,12 +452,26 @@ protected boolean initialize() {
...
             ok = stop = true;
+        } else {
+            // check restart is required or useless
+            if (M.getStrategy().isDynamic() != (restarter == AbstractRestart.NO_RESTART)) {
+                if (restarter == AbstractRestart.NO_RESTART) {
+                    if (getModel().getSettings().warnUser()) {
+                        logger.white().println("Combining static strategy with restarts is not recommended.");
+                    }
+                } else {
+                    if (getModel().getSettings().warnUser()) {
```
jgFages, line 470: "Using" instead of "Combining"

solver/src/main/java/org/chocosolver/solver/search/loop/move/IMoveFactory.java
```diff
@@ -96,58 +94,91 @@ default void setRestarts(LongCriterion restartCriterion, ICutoffStrategy restart
      * thanks to <code>restartStrategy</code>.
```
jgFages, line 94: Update doc, remove reference to Move

solver/src/main/java/org/chocosolver/solver/search/loop/move/IMoveFactory.java
```diff
@@ -96,58 +94,91 @@ default void setRestarts(LongCriterion restartCriterion, ICutoffStrategy restart
...
+     * @param restartLimit         restart limits (limit of number of restarts)
+     * @implNote Some counter may require an argument, as limit.
+     * This declared limit will be ignored and replaced by <i>base</i> on creation of this.
+     */
+    default void setConstantRestart(long scaleFactor, ICounter restartStrategyLimit, int restartLimit) {
+        ref().setRestarts(restartStrategyLimit, new MonotonicCutoff(scaleFactor), restartLimit);
     }
 
     /**
      * Creates a Move object that encapsulates the current move within a restart move.
```
jgFages, line 176: Update doc remove reference to Move

solver/src/main/java/org/chocosolver/solver/search/restart/AbstractRestart.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public static final AbstractRestart NO_RESTART = new AbstractRestart() {
+    };
+
+    public void init() {
+    }
+
+    public void setNext(AbstractRestart restart) {
+        throw new UnsupportedOperationException("Cannot set next on this instance");
+    }
```
jgFages, line 32: Why not putting the nextrestart variable in the abstract class?

solver/src/main/java/org/chocosolver/solver/search/restart/Restarter.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    private long limit;
+    /**
+     * Number of solutions found so far
+     */
+    private long solutions;
+
+    private final boolean resetCutoffOnSolution;
+
+    private AbstractRestart next = AbstractRestart.NO_RESTART;
```
jgFages, line 54: I would expect it to be in the abstract class (cf previous comment)

solver/src/main/java/org/chocosolver/solver/search/restart/Restarter.java
```diff
@@ -0,0 +1,113 @@
...
+ * This enables restarting a search on certain conditions
+ * (most of the time based on a counter).
+ * <p>
+ * Created by cprudhom on 03/09/15.
+ * Project: choco.
+ *
+ * @author Charles Prud'homme
+ * @since 03/09/2015
+ */
+public final class Restarter extends AbstractRestart {
```
jgFages, line 25: Is there a benefit from declaring this class final?

solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/AbstractStrategy.java
```diff
@@ -39,6 +39,12 @@ public boolean init(){
         return true;
     }
 
+    /**
+     * Indicates if this strategy is learn-based may, for instance, require to be coupled with a restart strategy.
+     * @return <i>true</i> if this strategy is dynamic.
+     */
+    public abstract boolean isDynamic();
```
jgFages, line 46: I find the concept of dynamic search quite vague and possibly confusing. 
If it is really about restarts i would suggest a more explicit name, like isImprovedByRestarts. 
However, i am not 100% convinced by the idea of this feature, because it often depends on the problem.

solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/ConflictOrderingSearch.java
```diff
@@ -104,6 +104,11 @@ public boolean init() {
         return mainStrategy.init();
     }
 
+    @Override
+    public boolean isDynamic() {
+        return true;
```
jgFages, line 109: For instance here it is not so clear to me that we should recommend restarts

jgFages:
Hi, I find the refactoring of the restart a very good idea, because it is indeed sometimes hard to follow. 

What would the code look like to restart both on solutions and every 100 fails for instance?

I am just not so sure about the isDynamical function.
I feel it is not so easy to answer true or false to that question (exept for inputOrderLB). So i wonder if it is worth increasing the codebase, especially if it is just for a warning.

solver/src/main/java/org/chocosolver/solver/search/restart/AbstractRestart.java
cprudhom: You're right. 
Well, that was because `NO_RESTART` singleton must not declare `next`, but I can manage that differently.

solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/AbstractStrategy.java
cprudhom: Indeed, the choice of the term and the concept it encapsulates are debatable. 
However, it happens quite often (and this is notably the case of the default strategy 🤦) that the strategy/restarts couple is unsuitable. 

However, this can be the subject of another RP with a more in-depth discussion.

jgFages:
Thanks for your answers!

> > What would the code look like to restart both on solutions and every 100 fails for instance?
> 
> ```java
> solver.setLubyRestart(100, new FailCounter(target, 0), 5000);
> solver.setRestartOnSolutions();
> ```
Before we had ```java ref().setMove(new MoveRestart(ref().getMove()``` , so the "set" was actually an add operation -> we had in the end the two restart strategies.
  
However, the new implementation looks like a real setter (the new Restarter will replace the previous one). 
It is more consistent with the method name, but I feel the luby restart will be erased by the setRestartOnSolutions.
I would expect the two restarts to be linked by a call to setNext()... Maybe I missed something

cprudhom:
True, that was my first intention but I didn't go through with it
I'll fix that. 
I can do that with a factory, which may ease the definition of the restarter (and we can keep the `setRestarter` method as is) and hides implementation details.

cprudhom:
Proposal (to be completed):

Before
```java
model.getSolver().setLubyRestart(2, new FailCounter(model, 2), 25000);
```

After
```java
model.getSolver().makeRestarter().luby(2).onFailure().upTo(25000);
```

jgFages (approved):

PR #956, opened 2022-10-22
Add variable set scope to search to make strategy sequencer faster

jgFages:
Should accelerate the usage of strategysequencer

I let the redundant V[] attribute to reduce impact on code and for compatibility purpose. However, if memory is an issue, a LinkedHashSet could replace the array and the set.

### DIFF
--- a/parsers/src/main/java/org/chocosolver/parser/RegParser.java
+++ b/parsers/src/main/java/org/chocosolver/parser/RegParser.java
@@ -212,8 +212,7 @@ public abstract class RegParser implements IParser {
         Solver solver = m.getSolver();
         if (solver.getSearch() != null) {
             THashSet<Variable> dvars = new THashSet<>();
-            dvars.addAll(Arrays.asList(solver.getSearch().getVariables()));
-            int k = 0;
+            dvars.addAll(solver.getSearch().getScope());
             IntVar[] ivars = m.streamVars()
                     .filter(VariableUtils::isInt)
                     .filter(v -> !VariableUtils.isConstant(v))
--- a/solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/AbstractStrategy.java
+++ b/solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/AbstractStrategy.java
@@ -29,7 +29,7 @@ import java.util.Set;
 public abstract class AbstractStrategy<V extends Variable>  {
 
     protected final V[] vars;
-    protected Set<V> scope;
+    protected final Set<V> scope;
 
     @SafeVarargs
     protected AbstractStrategy(V... variables) {
--- a/solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/AbstractStrategy.java
+++ b/solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/AbstractStrategy.java
@@ -15,6 +15,10 @@ import org.chocosolver.solver.search.strategy.decision.IntDecision;
 import org.chocosolver.solver.variables.IntVar;
 import org.chocosolver.solver.variables.Variable;
 
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
 /**
  * A search strategy provides decisions to go down in the search space.
  * The main method is {@link #computeDecision(Variable)} which returns the next decision to apply.
@@ -25,10 +29,16 @@ import org.chocosolver.solver.variables.Variable;
 public abstract class AbstractStrategy<V extends Variable>  {
 
     protected final V[] vars;
+    protected Set<V> scope;
 
     @SafeVarargs
     protected AbstractStrategy(V... variables) {
         this.vars = variables.clone();
+        this.scope = new HashSet<>(Arrays.asList(variables));
+    }
+
+    public final boolean contains(V var) {
+        return scope.contains(var);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


cprudhom:
@jgFages I'm not sure this is a quick win because I think the overhead has more to do with the type of complementary strategy.
As a recall, the complementary search (for `IntVar`) is :
```java
if (ivars.length > 0) {
    IntValueSelector valueSelector;
    if (m.getResolutionPolicy() == ResolutionPolicy.SATISFACTION
            || !(m.getObjective() instanceof IntVar)) {
        valueSelector = new IntDomainMin();
    } else {
        valueSelector = new IntDomainBest();
        m.getSolver().attach(m.getSolver().defaultSolution());
        valueSelector = new IntDomainLast(m.getSolver().defaultSolution(), valueSelector, null);
    }
    strats.add(Search.lastConflict(new IntStrategy(ivars, new DomOverWDegRef<>(ivars, 0), valueSelector)));
}
```

I profiled a minizinc instance (`roster-sickness`+ `large-2` from 2022) with the 2 alternatives:
- `-ocs OUTPUT` (default): 7056851 nodes in 110 seconds
- `-ocs NO`: 7056851 nodes in 81 seconds 
So clearly, the complementary search is useless.

With the first option, the profiler states that (filter on `org.chocosolver.solver.search.strategy.strategy`):
Capture d’écran 2022-10-24 à 13 10 45.
On the other hand, when I watch what's going on on conflict (*ie*, related to `DomWdeg`):
Capture d’écran 2022-10-24 à 13 23 58

So the question is double:
- Is a complementary search is required?
- if true, what is the right one to apply?

Or: Can an unnecessary complementary search be deactivated on the fly?

jgFages:
You are probably right. 

Regarding the need for a complementary search, i think we already checked and the answer was yes on some instances. 

However, using domoverwdeg for that sounds like an overkill. We should probably use mindomain or even input order instead, no?

cprudhom:
If I set the default strategy to `strats.add(Search.lastConflict(new IntStrategy(ivars, new FirstFail(m), valueSelector)));`, on this instance, I got the same numbers of nodes, but solve time is 84 seconds.

Now, I just need to track down the instances that were better solved with `Dom/WDeg`...

jgFages:
I think we established :
- that a complementary search was necessary because on some instances, we got uninstantiated vars after applying the input file search (maybe this is no longer true by the way).
- that domwdeg was a good option for the free/portfolio search

I am not sure we ever investigated the best complementary search when using the input file search...

cprudhom:
> I am not sure we ever investigated the best complementary search when using the input file search...

Actually, I did last year because of the poor results on some instances in MiniZinc 2021.

In fact, I had exchanged some emails with Jason Nguyen (from MiniZinc team):
I asked him:
> I have a question concerning the attitude to adopt, in the context of the competition, concerning models where the strategy is not complete, i.e. it does not allow to fix all the variables by simple propagation.
On ATSP+instance10_0p25, for instance, if I change the complementary search to only output variables (instead of all variables of the model), I'm able to find a better value in few seconds than the best one found by choco in the 2021 Challenge (that is, 4657236 in 130 sec).
> Is there any rules or advices ? 

He told me:
> It seems there were a few models this year which were not very well suited to the FD category.
> Ideally, models are all supposed to have a complete search strategy. Unfortunately, with much of our time going towards setting up new infrastructure to run the Challenge, we didn't have time to thoroughly check some models which were submitted late in the process.
> Generally, if you are given a problem without a complete search strategy, you are free to use any complementary search strategy. However, for future Challenges we will have some automated tooling to ensure that model submissions specify their search strategy correctly to hopefully avoid this issue altogether.

So maybe a very simple strategy would do the job.

About `ATSP`+`instance10_0p25` (from 2021) with 500s TL, best solution is :
- `NO`: throw exception `objective variable (X_INTRODUCED_315_ = [4807158,19806030]) is not instantiated on solution`
- `OUTPUT` (with 2021 default search): MINIMIZE X_INTRODUCED_315_ = 4708610, Resolution time 4,712s,
- `OUTPUT` (with DomWdeg, 2022 default search): MINIMIZE X_INTRODUCED_315_ = 4663656, Resolution time 426,043s
- `OUTPUT` (with FirstFail): MINIMIZE X_INTRODUCED_315_ = 4663656, Resolution time 314,705s

So, there was a need for a better complementary search, but I made an overkill choice. 
I still have to find the benchmark I used, but I probably deleted it...

jgFages:
Interesting! From his answer I understand that we can, but shouldn't have to complete it. 
So, if they ensure it next year, we probably do not need to waste more time on it.
I would just complete with inputLB/UB on objective variable only for min/max problems

Just to make sure I understand this: 
NO: throw exception objective variable (X_INTRODUCED_315_ = [4807158,19806030]) is not instantiated on solution
=> this corresponds to an error message on the first solution right? Not the best (you said you found better without complementary search and this bound seems worse than other solutions)?

cprudhom:
> Interesting! From his answer I understand that we can, but shouldn't have to complete it. So, if they ensure it next year, we probably do not need to waste more time on it. I would just complete with inputLB/UB on objective variable only for min/max problems

I think it is safer to complete with minDom+LB on output variables and potentially one on the objective variable if any.

> Just to make sure I understand this: NO: throw exception objective variable (X_INTRODUCED_315_ = [4807158,19806030]) is not instantiated on solution => this corresponds to an error message on the first solution right? Not the best (you said you found better without complementary search and this bound seems worse than other solutions)?

When the declared strategy cannot provide decisions, the problem is supposed to be fully instantiated which is not the case here, and an exception is thrown. So, yes this error happens on the first "solution".

cprudhom:
I propose to close the PR, #958 seems sufficient. Do you agree?

PR #958, opened 2022-10-26
Change the MiniZinc complementary search to a FirstFail

cprudhom:
This PR corrects the problem presented in #956 in a simpler way.
I also bet that the `mzn` test suite will be faster...🤞


jgFages (approved):

cprudhom:
Here again, tests are OK in my computer.
There is something wrong with the GH

PR #964, opened 2022-11-15
Dev memo

cprudhom:
It appears that PR #889 has introduced a memory consumption issue. 
Actually, the fact that propagators are not stored in an array (in `AbstractVariable`) but in an array of arrays requires too much memory.
Here are the memory consumption (in GB), estimated with `Runtime` and validated with JProfiler, of dummy models only composed of a very big number (1000000) of variables but no constraint:

| Version | IntVar | BoolVar | SetVar | GraphVar | 
| ----------- | ----------- |----------- |----------- |----------- |
| 4.10.8 | 0.50 | 0.41 | 1.28 | 0.79 |
| 4.10.9 | 1.46 | 0.66 | 1.83 | 3.5 |
| 4.10.11* | 0.39 | 0.30 | 1.16 | 0.69 |

Where `4.10.11*` denotes the proposal from this PR.
The targeted class is `AbstractVariable`, and the idea is mainly to create structures only when required.
The current way of storing a variable's propagators depends on the events they react on.
In many cases, all events are not really covered so there is no need to create everything all the time.

I also had a look to other array creation (in solver/*, ignoring /constraints/* directory first).


solver/src/main/java/org/chocosolver/solver/variables/impl/AbstractVariable.java
```diff
@@ -519,7 +522,47 @@ public ICause getCause() {
         return cause;
     }
 
-    static class BipartiteList {
+    static abstract class ABipartiteList {
```
jgFages, line 525: Suggestions : 
- externalize those classes to reduce file size
- it seems the abstract class has no field and could then be an interface with default implementation (the default implem could also be done as a separate class).

solver/src/main/java/org/chocosolver/solver/variables/impl/AbstractVariable.java
```diff
@@ -73,6 +73,12 @@ public abstract class AbstractVariable implements Variable {
      */
     static final String MSG_BOUND = "new bounds are incorrect";
 
+    /**
+     * Empty bipartite list
+     */
+    static final ABipartiteList EMPTY = new ABipartiteList() {
```
jgFages, line 79: in the end the JVM will create an anonymous class, maybe we could have a proper class for the default case

jgFages (dismissed):

cprudhom (commented):
Everything is ok to me

PR #966, opened 2022-11-16
Fix issue relative to incomplete filtering of atLeastNValues and atMostNValues

cprudhom:
The two constraints `model.atMostNValues()` and `model.atLeastNValues()` do not filter correctly the `n` variable.
Actually, `atMostNValues`only sharps the lower bound and `atLeastNValues` the upper bound.
So they can produce inconsistent solutions like:
```
6  = atMostNValues([0,0,0,0,0])
2 = atLeastNValues([1,2,3,4,5])
```

This PR fixes this issue.

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropAtLeastNValues.java
+++ b/solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropAtLeastNValues.java
@@ -95,7 +95,7 @@ public class PropAtLeastNValues extends Propagator<IntVar> {
             }
         }
         // filtering cardinality variable
-        vars[n].updateUpperBound(countMax, this);
+        vars[n].updateBounds(count, countMax, this);
         // filtering decision variables
         boolean again = false;
         if (count < countMax && countMax == vars[n].getLB()) {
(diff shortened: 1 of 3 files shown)


jgFages:
Unless I misunderstood something, I disagree, atMostNValue and atLeastNValues are not supposed to be sharp, so those solutions are consistent : 
6  = atMostNValues([0,0,0,0,0]) if the requirement is to have at most 6 values then 0,0,0,0 (only one value) is perfecty fine. And the filtering is not supposed to reduce the ub from 6 to 1.
It is the same for the atLeastNValues.

If one wants to compute sharp bounds, then the "nvalues" constraint should be used instead.

cprudhom:
Hmmm. I think I understand what I did wrong.
I was referring to the [global constraint catalog](https://sofdem.github.io/gccat/gccat/index.html) to check the right definition.
However the name used in choco is not consistent with the one in the catalog.
In choco, we use `atMostNValue` which refers to `max_value` in the [catalog](https://sofdem.github.io/gccat/gccat/Catmost_nvalue.html#uid14849) but I was looking for `atmost` in the [catalog](https://sofdem.github.io/gccat/gccat/Catmost.html).
So you are right.

The only confusing thing is that `N`might not be fixe by propagation.

PR #969, opened 2022-11-25
Reduce object creation during propagation

cprudhom:
In order to reduce the duration of the "mzn" test suite , because it is very unstable, I investigate objects creation during propagation and found out 2 memory leaks.
Once corrected, the `testThemAll` tests went from 5m30 to 4m30 on my machine (decrease of about 18%).

@fhermeni I also fixed a leak that you noticed.


solver/src/main/java/org/chocosolver/solver/constraints/real/PropScalarMixed.java
```diff
@@ -447,26 +456,17 @@ protected ESat check(double sumLB, double sumUB) {
                 if (sumLB <= b) {
                     return ESat.TRUE;
                 }
-                if (sumLB > b) {
```
jgFages, line 450: I would return false if sumUB > b and UNDEFINED otherwise, no?

jgFages: Yes like this it is now correct

jgFages (approved):

PR #985, opened 2022-12-07
Reduce object creation during solving

cprudhom:


jgFages:
Would you say more generally that one should avoid : 
- declaring lambda on the fly ?
- using streams ? 
when the calls are used at each node?

cprudhom:
Using stream during the resolution should be carefully done. 
If an alternative exists without stream, it should be preferred. Note, that it could require the introduction of new method (like I did with `Variable.forEachPropagator`).

Using lambdas is more or less the same.
If predefined functions/consumers/... can be extracted, that's better: profiling is easier and it limits the creation of objects (but it might increase the size of a model...).
Using lambda to save`IOperation` is OK, though, because a new instance is necessary.

jgFages:
Thanks. It confirms my experience of streams within algorithms.

PR #1007, opened 2023-02-14
Make possible to add labels to constraints and variables

cprudhom:
Nothing more than the title says ^^

The idea is to make possible to attach one or more `String` as labels for variables or contraints.
This can be useful to declare search strategies based on labeled variables, using streams.
Alternatively, this could be helpful to tag variables and constraints not created by the user.

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/Ilabelable.java
+++ b/solver/src/main/java/org/chocosolver/solver/Ilabelable.java
@@ -1,0 +1,43 @@
+/*
+ * This file is part of choco-solver, http://choco-solver.org/
+ *
+ * Copyright (c) 2023, IMT Atlantique. All rights reserved.
+ *
+ * Licensed under the BSD 4-clause license.
+ *
+ * See LICENSE file in the project root for full license information.
+ */
+package org.chocosolver.solver;
+
+/**
+ * An interface to allow adding labels to variables or constraints.
+ *
+ * @author Charles Prud'homme
+ * @since 10/02/2023
+ */
+public interface Ilabelable {
+    /**
+     * Add a label
+     *
+     * @param label to add
+     */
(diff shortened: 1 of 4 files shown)


jgFages:
I guess I lack context but I have some doubts regarding the usefulness of this PR :

Labelling variables : I understand it can be useful for declaring search. However, why not relying on the already existing name attribute ?

Labelling constraints : 
- would the idea to let the user label its constraints to understand that the remaining ones are built by the solver ? 
-> creating a set in the model would do the job in one line of code, no ?
- or would it be the opposite (labelling solver created constraints) ? 
-> Then I would be worried that has an impact on memory consumption. 

Maybe it could be more convincing with some examples ?

cprudhom:
You are probably right, there already exists such feature using maps or hooks in the model.
The first intention was to distinct, for challenges, decision variables from additional ones, but I can do that using hooks.

dimitri-justeau:
Hi ! Sorry to arrive like a "cheveu sur la soupe" on this closed PR, but about that I am wondering:

- How can hooks be used to tag variables as the decision variables ?

I think it could be useful to clearly distinct decision variables from other ones, notably for solution retrieval, or (my current issue) when using a ParallelPortfolio. Indeed, if I want to use the default strategies from the parallel portfolio (which are great) but only on decision variables, I need to rewrite a lot of codes. Also, if my model contains graph variables, I just cannot use the ParallelPortfolio with default strategies.

But maybe there is already a way to do that ?

jgFages:
Yes, hooks are a great feature! 
I do use them quite often actually

jgFages:
Hooks can be used to store anything in a model, including a set variables that you can call decision variables. 

Then filling this set and using it is up to you (as adding labels and using them)

PR #1011, opened 2023-02-17
Fix bug in removeAllValuesBut and removeValues and for enumerated domain variables

cprudhom:
I had this pernicious bug that occurs when the `removeAllValuesBut` method is called from  for enumerated domain variables. This occurs when the variable is also monitored (here, a task) : the bounds was not up to date. Which led to inconcistency.
To fix the bug, I encapsulate the internal `updateBounds` call in a loop to make sure that side effects (monitors or views) are capture at the right time.
I also fix a bug in removeValues.

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetArrayIntVarImpl.java
+++ b/solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetArrayIntVarImpl.java
@@ -275,11 +275,12 @@ public final class BitsetArrayIntVarImpl extends AbstractVariable implements Int
             hasChanged |= fixpoint;
         } while (fixpoint);
         // now deal with holes
+        int to = UB.get() - 1;
         boolean hasRemoved = false;
         int count = SIZE.get();
         int value;
         // iterate over the values in the domain, remove the ones that are not in values
-        for (int index = INDICES.nextSetBit(nlb + 1); index > -1 && index <= nub; index = INDICES.nextSetBit(index + 1)) {
+        for (int index = INDICES.nextSetBit(LB.get() + 1); index > -1 && index <= to; index = INDICES.nextSetBit(index + 1)) {
             value = VALUES[index];
             if (!values.contains(value)) {
                 model.getSolver().getEventObserver().removeValue(this, value, cause);
--- a/solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetArrayIntVarImpl.java
+++ b/solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetArrayIntVarImpl.java
@@ -191,30 +191,35 @@ public final class BitsetArrayIntVarImpl extends AbstractVariable implements Int
     @Override
     public boolean removeValues(IntIterableSet values, ICause cause) throws ContradictionException {
         assert cause != null;
-        int olb = getLB();
-        int oub = getUB();
-        int nlb = values.nextValue(olb - 1);
-        int nub = values.previousValue(oub + 1);
-        if (nlb > oub || nub < olb) {
-            return false;
-        }
-        int i;
-        // look for the new lb
-        while (nlb == olb && olb < Integer.MAX_VALUE) {
-            i = INDICES.nextSetBit(V2I.get(olb) + 1);
-            olb = i > -1 ? VALUES[i] : Integer.MAX_VALUE;
+        boolean hasChanged = false, fixpoint;
+        int nlb, nub;
+        do {
+            int olb = getLB();
+            int oub = getUB();
             nlb = values.nextValue(olb - 1);
-        }
-        if (nlb <= nub) {
-            // look for the new ub
-            while (nub == oub && oub > Integer.MIN_VALUE) {
-                i = INDICES.prevSetBit(V2I.get(oub) - 1);
-                oub = i > -1 ? VALUES[i] : Integer.MIN_VALUE;
-                nub = values.previousValue(oub + 1);
+            nub = values.previousValue(oub + 1);
+            if (!hasChanged && (nlb > oub || nub < olb)) {
+                return false;
             }
-        }
-        // the new bounds are now known, delegate to the right method
-        boolean hasChanged = updateBounds(olb, oub, cause);
+            int i;
+            // look for the new lb
+            while (nlb == olb && olb < Integer.MAX_VALUE) {
+                i = INDICES.nextSetBit(V2I.get(olb) + 1);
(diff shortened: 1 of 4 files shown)


solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetArrayIntVarImpl.java
```diff
@@ -191,30 +191,35 @@ public boolean removeValue(int value, ICause cause) throws ContradictionExceptio
...
+                while (nub == oub && oub > Integer.MIN_VALUE) {
+                    i = INDICES.prevSetBit(V2I.get(oub) - 1);
+                    oub = i > -1 ? VALUES[i] : Integer.MIN_VALUE;
+                    nub = values.previousValue(oub + 1);
+                }
+            }
+            // the new bounds are now known, delegate to the right method
+            fixpoint = updateBounds(olb, oub, cause);
+            hasChanged |= fixpoint;
+        }while(fixpoint);
```
ArthurGodet, line 222: This line is not formatted as other ones.

solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetArrayIntVarImpl.java
```diff
@@ -244,27 +249,32 @@ public boolean removeValues(IntIterableSet values, ICause cause) throws Contradi
...
+                i = INDICES.nextSetBit(V2I.get(olb) + 1);
+                olb = i > -1 ? VALUES[i] : Integer.MAX_VALUE;
+                nlb = values.nextValue(olb - 1);
             }
-        }
-        // the new bounds are now known, delegate to the right method
-        boolean hasChanged = updateBounds(nlb, nub, cause);
+            if (nlb <= nub) {
+                // look for the new ub
+                while (nub != oub && olb > Integer.MIN_VALUE && oub > Integer.MIN_VALUE) {
```
ArthurGodet, line 268: Shouldn't the test be on nub instead of olb ?

solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetIntVarImpl.java
```diff
@@ -191,30 +191,35 @@ public boolean removeValue(int value, ICause cause) throws ContradictionExceptio
...
+                while (nub == oub && oub > Integer.MIN_VALUE) {
+                    i = VALUES.prevSetBit(nub - 1 - OFFSET);
+                    oub = i > -1 ? i + OFFSET : Integer.MIN_VALUE;
+                    nub = values.previousValue(oub + 1);
+                }
+            }
+            // the new bounds are now known, delegate to the right method
+            fixpoint = updateBounds(olb, oub, cause);
+            hasChanged |= fixpoint;
+        }while(fixpoint);
```
ArthurGodet, line 222: Line not formatted as other ones.

solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetArrayIntVarImpl.java
```diff
@@ -191,30 +191,35 @@ public boolean removeValue(int value, ICause cause) throws ContradictionExceptio
...
+            if (nlb <= nub) {
+                // look for the new ub
+                while (nub == oub && oub > Integer.MIN_VALUE) {
+                    i = INDICES.prevSetBit(V2I.get(oub) - 1);
+                    oub = i > -1 ? VALUES[i] : Integer.MIN_VALUE;
+                    nub = values.previousValue(oub + 1);
+                }
+            }
+            // the new bounds are now known, delegate to the right method
+            fixpoint = updateBounds(olb, oub, cause);
```
ArthurGodet, line 220: I think that it comes from finishing conditions of the algorithm, but we are using "old" bounds variables her instead of new ones (nlb and nub). It might be clearer to call them.

solver/src/main/java/org/chocosolver/solver/variables/impl/BitsetIntVarImpl.java
```diff
@@ -191,30 +191,35 @@ public boolean removeValue(int value, ICause cause) throws ContradictionExceptio
...
+            if (nlb <= nub) {
+                // look for the new ub
+                while (nub == oub && oub > Integer.MIN_VALUE) {
+                    i = VALUES.prevSetBit(nub - 1 - OFFSET);
+                    oub = i > -1 ? i + OFFSET : Integer.MIN_VALUE;
+                    nub = values.previousValue(oub + 1);
+                }
+            }
+            // the new bounds are now known, delegate to the right method
+            fixpoint = updateBounds(olb, oub, cause);
```
ArthurGodet, line 220: I think that it comes from finishing conditions of the algorithm, but we are using "old" bounds variables her instead of new ones (nlb and nub). It might be clearer to call them.

solver/src/main/java/org/chocosolver/solver/variables/view/IntView.java
```diff
@@ -167,29 +167,34 @@ public boolean removeValue(int value, ICause cause) throws ContradictionExceptio
...
+            }
+            if (nub == oub) {
+                // look for the new ub
+                do {
+                    oub = previousValue(oub);
+                    nub = values.previousValue(oub + 1);
+                } while (olb > Integer.MIN_VALUE && oub > Integer.MIN_VALUE && nub == oub);
+            }
+            // the new bounds are now known, delegate to the right method
+            fixpoint = updateBounds(olb, oub, cause);
```
ArthurGodet, line 195: I think that it comes from finishing conditions of the algorithm, but we are using "old" bounds variables her instead of new ones (nlb and nub). It might be clearer to call them.

ArthurGodet (dismissed):

cprudhom:
Now that I look back to the code, I think there is a better alternative to while-loops.
I will suggest another correction soon

cprudhom:
> Now that I look back to the code, I think there is a better alternative to while-loops. I will suggest another correction soon

I finally let the code as is, I don't want the code to introduce to much code

ArthurGodet (commented):
What the v and the c stand for in vlb/vub and clb/cub ?
Otherwise, seems fine to me !

cprudhom:
> What the v and the c stand for in vlb/vub and clb/cub ? Otherwise, seems fine to me !

I am not very inspired: `v` stands for `values`, name of the parameter in the API, and `c` stands for `current`.

ArthurGodet (dismissed):

PR #1015, opened 2023-02-27
Fix #1014

### LINKED ISSUE #1014: [BUG] PropUnionVar has index errors
**Issues**
- PropUnionVar may encounter ArrayIndexOutOfBoundsException

- The documentation of  
````default Constraint union(SetVar unionSet, int vOffset, SetVar indices, int iOffset, SetVar[] sets) ````
does not explain how the offsets are taken into account in the formula. My expectation would be the following :
<i>sets[i-iOffset]-vOffset</i>, where <i>i</i> in <i>indices</i>, is equal to <i>unionSet</i>.
but I do not know if this is was indeed the intended implementation.

- Finally, even though I am not 100% sure about the vOffset exact definition, I think that it should somehow intervene in ````filter1 and filter2````, whereas it is not the case.  

**To Reproduce**

```
@Test(groups = "1s", timeOut = 60000)
public void testUnionBounds() {
    Model model = new Model();
    SetVar[] sets = model.setVarArray("S", 3, new int[]{}, new int[]{2, 3, 5, 6});
    SetVar indices = model.setVar("I", new int[]{}, new int[]{-1, 0, 2, 6});
    SetVar union = model.setVar("U", new int[]{}, new int[]{1, 2, 3, 6, 7});
    model.union(union, indices, sets).post();
    model.getSolver().solve();
}
```

```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 5
	at org.chocosolver.solver/org.chocosolver.solver.constraints.set.PropUnionVar.filter2(PropUnionVar.java:127)
```
	
	
This time with the following code : (array of 2 sets but index set = {0, 1, 2}, I would expect to have no solution as you cannot have 3 indexes for an array of size 2. Instead, the test outputs a solution.

```
    @Test(groups = "1s", timeOut = 60000)
    public void testUnionWrongBoundsResult() {
        Model model = new Model();
        SetVar[] sets = model.setVarArray("S", 2, new int[]{}, new int[]{0,1,2});
        SetVar indices = model.setVar("I", new int[]{}, new int[]{0,1,2});
        SetVar union = model.setVar("U", new int[]{}, new int[]{0,1});
        model.union(union, indices, sets).post();
        model.getSolver().solve();
        model.getSolver().printStatistics();
    }
```

gives a solution (whereas it should not). Here vars[2] is actually returning the union variable

```
    @Test(groups = "1s", timeOut = 60000)
    public void testUnionWrongBoundsResult() {
        Model model = new Model();
        SetVar[] sets = model.setVarArray("S", 2, new int[]{}, new int[]{0,1,2});
        SetVar indices = model.setVar("I", new int[]{0,1,2,3,4,5}, new int[]{0,1,2,3,4,5});
        SetVar union = model.setVar("U", new int[]{}, new int[]{0,1});
        model.union(union, indices, sets).post();
        model.getSolver().solve();
        model.getSolver().printStatistics();
    }
```

gives a solution (whereas it should not). Here the index set is somehow ignored as it is instantiated.
    
```
    @Test(groups = "1s", timeOut = 60000)
    public void testUnionWrongBoundsResult() {
        Model model = new Model();
        SetVar[] sets = model.setVarArray("S", 2, new int[]{}, new int[]{0,1,2});
        SetVar indices = model.setVar("I", new int[]{}, new int[]{0,1,2,3,4});
        SetVar union = model.setVar("U", new int[]{}, new int[]{0,1});
        model.union(union, indices, sets).post();
        model.getSolver().solve();
        model.getSolver().printStatistics();
    }
```

runs with error... but it would be better to have no solution than a runtime exception.

**Possible solution**

It will not solve everything but I suggest to remove vOffset to make the code clearer. I think this argument is not necessary as we already have an offset constraint if we really want to shift a set variable (so this can be done outside of this constraint).

The domain of the index set should be filtered according to sets.length and iOffset.

**Environment (please complete the following information):**
- Choco-solver version: 4.10.12-SNAPSHOT
- JRE : 11

### PR DESCRIPTION
**Pull request by cprudhom**

### CODE REVIEW

File: solver/src/main/java/org/chocosolver/solver/constraints/set/PropUnionVar.java [hunk 1]
```diff
@@ -70,6 +68,20 @@ public PropUnionVar(SetVar union, int vOffset, SetVar indices, int iOffset, SetV
     public void propagate(int evtmask) throws ContradictionException {
         SetVar union = vars[k];
         SetVar indices = vars[k + 1];
+        if (indices.getUB().size() > 0) {
```
jgFages, line 71: I think this filtering should be called only once, maybe embedding it in a separate method called depending on the evtmask or an internal variable ?

solver/src/main/java/org/chocosolver/solver/constraints/set/PropUnionVar.java
```diff
@@ -70,6 +68,20 @@ public PropUnionVar(SetVar union, int vOffset, SetVar indices, int iOffset, SetV
...
+            int min = indices.getUB().min();
+            while (min - iOffset < 0 && indices.getUB().size() > 0) {
+                indices.remove(min, this);
+                min = indices.getUB().min();
+            }
+        }
+        if (indices.getUB().size() > 0) {
+            int max = indices.getUB().max();
+            while (max - iOffset > k && indices.getUB().size() > 0) {
```
jgFages, line 80: >= k ?

solver/src/main/java/org/chocosolver/solver/constraints/set/PropUnionVar.java
cprudhom: I can do that, indeed

solver/src/main/java/org/chocosolver/solver/constraints/set/PropUnionVar.java
cprudhom: `k` is equal to sets.length I'll change that

cprudhom: I won't change, this is a class variable

solver/src/main/java/org/chocosolver/solver/constraints/set/PropUnionVar.java
```diff
@@ -68,14 +67,36 @@ public PropUnionVar(SetVar union, int vOffset, SetVar indices, int iOffset, SetV
     //***********************************************************************************
     @Override
     public void propagate(int evtmask) throws ContradictionException {
-        SetVar union = vars[k];
         SetVar indices = vars[k + 1];
+        if (firstProp) {
+            firstProp = false;
+            model.getEnvironment().save(() -> firstProp = true);
```
jgFages, line 73: For my knowledge is there a big difference between this and using an IStateBool ?

jgFages (approved):

solver/src/main/java/org/chocosolver/solver/constraints/set/PropUnionVar.java
cprudhom: Ask yourself :)
The main interest of `IOperation` is that it is GC on backtrack where `IStateBool` is not.
Otherwise, it is comparable

Issue #1045, opened 2023-06-26
[BUG] fzn-choco, table, wrong answer


matsc-at-sics-se:
Choco 4.10.13.

Choco reports three solutions on this one: `A=1, A=7, A=18`. `A=1` is wrong.
```
include "globals.mzn";
var {1,2,4,7,18}: A;
constraint table([A,A,], [| 1, 4, | 2, 1, | 4, 7, | 7, 2, | 7, 7, | 18, 18, |]);
solve :: int_search([A], largest, indomain_interval, complete) satisfy;
```

cprudhom:
When multiple occurrences of variables are in the scope of table constraint, some algorithms does not work as expected.
The solution is to replace such occurrences by views.

matsc-at-sics-se:
I pulled and recompiled yesterday. The wrong solution is still there. Just letting you know.

cprudhom:
Thank you, I had partly done the work.

Issue #1044, opened 2023-06-26
[BUG] fzn-choco, lex_less, wrong answer


matsc-at-sics-se:
Choco 4.10.13.

This succeeds with `A = 2`, which is wrong.
```
include "globals.mzn";
var 1..3: A;
constraint lex_less([A,0],[A,0]);
```

Issue #1043, opened 2023-06-26
[BUG] fzn-choco, mod, wrong solution


matsc-at-sics-se:
Choco 4.10.13.

Choco report `A = -1, B = -7` for this one, which is wrong.
```
include "globals.mzn";
var {-8,-1}: A;
var {-8,-7,-2}: B;
constraint B mod B=A;
```

cprudhom:
Related to #1045

Issue #1042, opened 2023-06-26
[BUG] fzn-choco, nvalue, wrong solution


matsc-at-sics-se:
Choco 4.10.13.

This model has one solution, `C = 2, D = 3`, but Choco also finds `C = D = 3`, which is wrong.
```
include "globals.mzn";
var 2..3: C;
var 3..4: D;
constraint nvalue(D,[1,D,C,1]);
solve :: int_search([C,D], input_order, indomain_min, complete) satisfy;
```

PR #1048, opened 2023-06-27
Uniquesafe : ensure correct behavior uniqueness of variables in scope is not satisfied

cprudhom:
This branch is related to #1040 #1041 #1042 #1043 #1044 and #1045.
The provided solution is to change the arguments of the propagators to add views instead of multiple occurrences of the same variables.
All constraints must be checked, but the methods added in `IIntConstraintFactory` are a first step.


solver/src/main/java/org/chocosolver/solver/constraints/reification/PropXltYCReif.java
```diff
@@ -43,8 +43,8 @@ public void propagate(int evtmask) throws ContradictionException {
                     this.setPassive();
                 }
             } else {
-                vars[0].updateLowerBound(vars[1].getLB() + cste, this);
-                vars[1].updateUpperBound(vars[0].getUB() - cste, this);
+                while(vars[0].updateLowerBound(vars[1].getLB() + cste, this) |
```
jgFages, line 46: is the while still necessary after using the views?
if it is used to handle the case where x=y, would it be possible to catch it before and simplify as
if (c > 0) then b = 1
else b = 0
?

cprudhom: Actually, that propagator can be safe regarding the non-uniqueness of variables by adding this line of code. 

I should have mentioned that there are 3 ways to fix issues related to multiple occurrences of a variable in a propagator:
1. ensuring that the propagator is indempotent
2. duplicating the propagator
3. using views

The first one should be favored but the the other ones are easy to setup.

jgFages: ok thanks, but regarding my second point. 
if the constraint is "b <=> x < x + c"
If c > 0, the constraint is always true so we should set b to 1 and set propagator to passive
Else we should set b to false and set propagator to passive

I believe it would make the code clearer and more efficient, no?

PR #1052, opened 2023-07-12
Improve PropNValue.contains().

fhermeni:
This used to run in O(n). Given the array is read-only, it is ok to sort it and use Arrays.binarySearch() to perform in O(log(n)).

For an internal benchmark with 128 values to watch, contains() was the bottleneck (5% of the service latency), now it is < 1%.


jgFages (approved):

jgFages:
Thanks Fabien !

PR #1054, opened 2023-07-21
use square when possible and stronger filtering

jgFages:
practical interest to be confirmed by experiments

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
+++ b/solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
@@ -117,7 +117,6 @@ public class PropSquare extends Propagator<IntVar> {
 
     protected void updateUpperBoundofX() throws ContradictionException {
         vars[0].updateUpperBound(Math.max(sqr(vars[1].getLB()), sqr(vars[1].getUB())), this);
-
     }
 
     protected boolean updateHolesinX() throws ContradictionException {
@@ -158,7 +157,11 @@ public class PropSquare extends Propagator<IntVar> {
     }
 
     protected boolean updateLowerBoundofY() throws ContradictionException {
-        return vars[1].updateLowerBound(-ceil_sqrt(vars[0].getUB()), this);
+        if (vars[1].getLB() >= 0) {
+            return vars[1].updateLowerBound(ceil_sqrt(vars[0].getLB()), this);
+        } else {
+            return vars[1].updateLowerBound(-floor_sqrt(vars[0].getUB()), this);
+        }
     }
 
     protected boolean updateUpperBoundofY() throws ContradictionException {
@@ -166,8 +169,19 @@ public class PropSquare extends Propagator<IntVar> {
     }
 
     protected boolean updateHolesinY() throws ContradictionException {
-        // remove intervals to deal with consecutive value removal and upper bound modification
-        if (bothEnum) {
+        if (vars[0].isInstantiatedTo(0)) {
+            return vars[1].instantiateTo(0, this);
+        }
+        boolean impact = false;
+        // remove interval around 0 based on X LB
+        if (vars[1].hasEnumeratedDomain()) {
+            int val = ceil_sqrt(vars[0].getLB()) - 1;
+            if (val >= 0) {
+                impact = vars[1].removeInterval(-val, val, this);
+            }
+        }
+        // remove values based on X holes
+        if (bothEnum && hasHoles(vars[0])) {
             int ub = vars[1].getUB();
             vrms.clear();
             vrms.setOffset(vars[1].getLB());
@@ -176,24 +190,12 @@ public class PropSquare extends Propagator<IntVar> {
                     vrms.add(value);
                 }
             }
-            return vars[1].removeValues(vrms, this);
-        } else if (vars[1].hasEnumeratedDomain()) {
-            int lb = vars[1].getLB();
-            int ub = vars[1].getUB();
-            while (!vars[0].contains(sqr(lb))) {
-                lb = vars[1].nextValue(lb);
-                if (lb > ub) break;
-            }
-            boolean filter = vars[1].updateLowerBound(lb, this);
-
-            while (!vars[0].contains(sqr(ub))) {
-                ub = vars[1].nextValue(ub);
-                if (ub < lb) break;
-            }
-            return filter | vars[1].updateUpperBound(ub, this);
+            impact |= vars[1].removeValues(vrms, this);
         }
-        return false;
+        return impact;
     }
 
-
+    private boolean hasHoles(IntVar var) {
+        return (var.getUB() - var.getLB() + 1) > var.getDomainSize();
+    }
 }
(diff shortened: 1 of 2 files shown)


ArthurGodet (commented):
Basically, your modifications are coorect, but I think we can go way further.
First, there is an error in the isEntailed() method.
Second, I have the impression that the filtering is not always consistent between X and Y and can be improved. For instance, at root node, and if X has an enumerated domain, why not remove all values that are not a perfect square ?
Third, the propagator of such a "simple" constraint should reacts to fine events, don't you think ?

It might look like a "total" revamp of the propagator, but I think it is worth it. What do you think ?

(I can help with the revamp if needed)

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
```diff
@@ -117,57 +108,95 @@ protected void updateLowerBoundofX() throws ContradictionException {
...
-                        (vars[1].contains(floor_sqrt(value)) || vars[1].contains(-floor_sqrt(value))))) {
+                int sqrt = floor_sqrt(value);
+                if (!vars[1].contains(sqrt) && !vars[1].contains(-sqrt)) {
                     vrms.add(value);
                 }
             }
-            return vars[0].removeValues(vrms, this);
-        } else if (vars[0].hasEnumeratedDomain()) {
+            impact |= vars[0].removeValues(vrms, this);
+        } else {
```
ArthurGodet, line 144: The "else" might be only if vars[1].hasEnumeratedDomain(), doesn't it ? If vars[0] has an enumerated domain (and therefore vars[1] has not), this part of code will never do anything (since there are no two integers i and j such that i^2 == j^2 - 1, except 0 and 1) which is not done in updateLowerBoundofX() or updateUpperBoundofX().

Moreover, inside the while loops, value should be updated with the next (or previous) perfect square value instead of vars[0].nextValue(value) (or vars[0].previousValue(value)) as these values won't have been removed at root node if vars[0] has not an enumerated domain. By the way, such methods should be used directly in updateLowerBoundofX() and updateUpperBoundofX().

That said, with these corrections, I don't see a situation where this "else" code is doing something (especially with the statement returning false if vars[0] has not an enumerated domain).

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
```diff
@@ -117,57 +108,95 @@ protected void updateLowerBoundofX() throws ContradictionException {
...
             int value = vars[0].getLB();
             int nlb = value - 1;
             while (nlb == value - 1) {
-                if (!vars[1].contains(floor_sqrt(value)) && !vars[1].contains(-floor_sqrt(value))) {
+                int sqrt = floor_sqrt(value);
```
ArthurGodet, line 148: This statement is valid if and only if value is a perfect square value (which is valid if and only if vars[0] has an enumerated domain). Otherwise, it should be `int sqrt = ceil_sqrt(value);`.

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
```diff
@@ -117,57 +108,95 @@ protected void updateLowerBoundofX() throws ContradictionException {
...
+            return vars[1].updateUpperBound(floor_sqrt(vars[0].getUB()), this);
+        }
     }
 
     protected boolean updateHolesinY() throws ContradictionException {
-        // remove intervals to deal with consecutive value removal and upper bound modification
-        if (bothEnum) {
+        if (vars[0].isInstantiatedTo(0)) {
+            return vars[1].instantiateTo(0, this);
+        }
```
ArthurGodet, line 189: Is this useful ? If vars[0] is instantiated to 0, vars[1] should already be valued to 0 with updateLowerBoundofY() and updateUpperBoundofY().

It might be more useful to have an if statement returning false if vars[1] has not an enumerated domain.

solver/src/test/java/org/chocosolver/solver/constraints/binary/SquareTest.java
```diff
@@ -23,6 +23,57 @@
...
+        IntVar[] x2 = m.intVarArray(n, -n*n,n*n, true);
+        for (int i=0;i<x.length;i++) {
+            m.square(x2[i], x[i]).post();
+        }
+        m.allDifferent(x).post();
+        m.allDifferent(x2).post();
+        while (m.getSolver().solve()) ;
+        assertEquals(m.getSolver().getSolutionCount(), 184320);
+        assertEquals(m.getSolver().getFailCount(), 209153);
+        assertEquals(m.getSolver().getNodeCount(), 577792);
```
ArthurGodet, line 40: This is too dependent from the default search. I completely see the interest of the tests (showing the strengthening of filtering with enumerated domains), but the tests might become false if we change the default search and I would rather prefer that they stay true in such a case.

ArthurGodet (dismissed):

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
jgFages: The else had a vars[0].hasEnumeratedDomain() condition that is now at the beginning of the methode.

Actually, I was thinking that maybe the else should be completely removed. I am not sure there is more to filter in such case. Isn't this filtering already ensured by the setBounds() ?

jgFages: If there is more to filter, could you please provide an exemple / test case in SquareTest ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
jgFages: I has been introduce in a commit to fix a bug so, since it has no impact on performance, I preferred to keep it.

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
ArthurGodet: It was my point (sorry if it was not clear). I think the else is useless whenever vars[1] has not an enumerated domain, as the setBounds() method should already have filtered the bounds.

However, there is one thing that should be added to the updateLowerBoundofX() and updateUpperBoundofX() methods: a check that the new bound is a perfect square (which is not the case if both variables haven't enumerated domains and y.LB <= 0 <= y.UB).

ArthurGodet:
I've simplified the code of updateHoles methods. More especially, I've removed dead code (which gave a welcome speed up, some tests taking only 30% of times compared to what they took before) and standardised the code between similar functions.

@jgFages If you are ok with modifications, I think we can merge the code.

jgFages:
@ArthurGodet the git history has been corrupted with merge commits (some from 2022), this would make history harder to analyze, so I close this PR and I create a new one here : https://github.com/chocoteam/choco-solver/pull/1055

I will push a last modification and comment your suggestions there

PR #1055, opened 2023-08-01
use square when possible and stronger filtering 2

jgFages:


solver/src/main/java/org/chocosolver/solver/constraints/binary/PropSquare.java
```diff
@@ -176,24 +205,12 @@ protected boolean updateHolesinY() throws ContradictionException {
...
+            impact |= vars[1].removeValues(vrms, this);
         }
-        return false;
+        return impact;
     }
 
-
+    private boolean hasHoles(IntVar var) {
+        return (var.getUB() - var.getLB() + 1) > var.getDomainSize();
+    }
```
jgFages, line 215: @ArthurGodet you suggest to remove it because the following code is expected to create holes everywhere?
```
// check perfect squares once and for all
            int ub = vars[0].getUB();
            vrms.clear();
            vrms.setOffset(vars[0].getLB());
            for (int value = vars[0].getLB(); value <= ub; value = vars[0].nextValue(value)) {
                if (!MathUtils.isPerfectSquare(value)) {
                    vrms.add(value);
                }
            }
            impact = vars[0].removeValues(vrms, this);
```

ArthurGodet: Yes, indeed. As long as vars[0] has an enumerated domain and its bounds are different, it will necessarily contains holes because of this code.

ArthurGodet:
Yes, everything seems good to me now !

ArthurGodet (approved):

PR #1057, opened 2023-08-21
Add frequency + fallback value selector

cprudhom:
There's no doubt that BIVS is very useful. However, it is also time-consuming and it is not always appropriate to use it.
I propose, in this PR, to make it possible to indicate an alternative frequency and strategy.


jgFages:
I agree that BIVS is costly. However, I would prefer the condition to be a generic function : IntVar -> Boolean. One implementation could be to be based on the number of backtracks, another could be based on the variable's name or domain. It could also be random...

cprudhom:
Ok, I can change to a more functional behavior even if sometimes when there's too much freedom you don't know what to choose :)

parsers/src/test/resources/xcsp/instances.csv
```diff
@@ -3,14 +3,14 @@ basics;Allergy.xml.lzma;1;_;1;0
 basics;AllInterval-005.xml.lzma;1;_;13;9
 basics;Auction-cnt-example_c18.xml.lzma;2;54;7;4
 basics;Auction-sum-example_c18.xml.lzma;2;54;6;3
-basics;Bacp-m1-06_c18.xml.lzma;3;10;38111;37700
```
jgFages, line 6: shouldn't we add the time ? 
It should be compared with care but still it would be nice to have an idea if it is faster or slower

cprudhom: Honestly, I won't add this for the reason described here (https://buildjet.com/for-github-actions/blog/why-github-actions-is-so-slow#what-hardware-is-actually-github-actions-using).
In two words,  you cannot select the hardware the VM is running on.
This explains why Minizinc tests often fail on GH even though they work on my computer.

It's something I've been wanting to check for a long time now, but I can't find the right way to do it.

jgFages: I agree that GitHub test runtime is hard to trust, but this file is updated locally (i.e. on your computer), no ?
Even if it is not perfect (various contributors may have several machines), it may be better than nothing...

jgFages (dismissed):

parsers/src/test/resources/xcsp/instances.csv
cprudhom: Such artifact already exists. For instance, https://github.com/chocoteam/choco-solver/actions/runs/5766875782 you can get the artifact which contains 2 files, one for xcsp and the other for mzn

cprudhom: Maybe the question should be: how to test performance regression?

cprudhom:
I propose to define the default application of BIVS each time, i.e. `v -> true` in order to keep the performance tests unchanged.

PR #1058, opened 2023-08-23
Introducing blackbox search configurator 

cprudhom:
This PR aims at introducing a class, namely `BlackBoxConfigurator`.
This should ease the way search is configured. 

Here is an bunch of code:
```java 
BlackBoxConfigurator bb = BlackBoxConfigurator.init();
bb.setIntVarStrategy(Search::failureRateBasedSearch)
  .setMetaStrategy(Search::lastConflict)
  .setRestartPolicy(solver -> AbstractRestart.NO_RESTART);
bb.make(model);
```
Doing so, it makes possible to pre-defined some configurations and applies them in one-line call.

It can appear to be redundant with `Search`. However, `Search` does not manage restarts or nogod-recording.
Here, in one class you have everything (hopefully) you should think of when it comes to define a search strategy.


solver/src/main/java/org/chocosolver/solver/search/strategy/BlackBoxConfigurator.java
```diff
@@ -0,0 +1,403 @@
...
+                .setIntObjVarStrategy((obj, max) -> max ? Search.minDomUBSearch(obj) : Search.minDomLBSearch(obj))
+                .setSetVarStrategy(Search::setVarSearch)
+                .setGraphVarSearch(Search::graphVarSearch)
+                .setRealVarStrategy(Search::realVarSearch)
+                .setRealObjVarStrategy((obj, max) -> Search.realVarSearch(new Cyclic<>(), max ? new RealDomainMax() : new RealDomainMin(),
+                        !max, obj))
+                .setMetaStrategy(Search::lastConflict)
+                .setRestartPolicy(solver -> AbstractRestart.NO_RESTART)
+                .setRestartOnSolution(false)
+                .setNogoodOnRestart(false)
```
ArthurGodet, line 133: Comments of the class indicate that it is valued to true by default. Either the comment or this value should me modified. And if set to true here, methods forCsp() and forCop() should be modified in order not to declare true a variable already set to true.

ArthurGodet (changes requested):

ArthurGodet (approved):
This is a great idea, it should really simplify the process of declaring searches.

PR #1059, opened 2023-08-29
Improve implementation of PropPowEven and PropPowOdd

ArthurGodet:
This pull request improves the implementation of PropPowEven and PropPowOdd from what was done on PropSquare. As this propagator was not doing anything more than what is done by PropPowEven, I deleted it, but kept the SquareTest class.


solver/src/test/java/org/chocosolver/solver/constraints/binary/SquareTest.java
```diff
@@ -43,8 +43,8 @@ public void testBigBoundBound() {
         solver.setSearch(Search.randomSearch(ArrayUtils.append(x,x2), 0));
         while (solver.solve()) ;
         assertEquals(solver.getSolutionCount(), 184320);
-        assertEquals(solver.getFailCount(), 412230);
-        assertEquals(solver.getNodeCount(), 780869);
+        assertEquals(solver.getFailCount(), 270905);
```
jgFages, line 46: why does the number of nodes change ?
which filtering rule was missing ? 
(It could also be that the new filtering is weaker and the search more lucky)

ArthurGodet: It was in method updateLowerBoundofX(). Whenever both variables had bounded domains and Y.getLB() < 0 < Y.getUB(), then we would not filter the lower of X even if it didn't correspond to a perfectNth value. In the case of square values, let say that X in [7,25] and Y in [-5,5]. With former code, X's lower bound would not have been updated to 9, whereas it is the case now.

jgFages: shouldn't we have a similar treatment in updateUpperBoundofX() ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropPowEven.java
```diff
@@ -91,14 +76,16 @@ void setBounds() throws ContradictionException {
     }
 
     int floor_nroot(int n) {
-        if (n < 0)
+        if (n < 0) {
             return 0;
+        }
         return MathUtils.safeCast((long) Math.floor(Math.pow(n, 1. / exponent)));
```
jgFages, line 82: I wonder if Math.sqrt(n) has not a more efficient implementation than Math.pow(n, 1. / 2)

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropPowEven.java
```diff
@@ -91,14 +76,16 @@ void setBounds() throws ContradictionException {
...
     }
```
jgFages, line 83: shouln't we use PRECISION here as well ?

solver/src/test/java/org/chocosolver/solver/constraints/binary/SquareTest.java
ArthurGodet: No, the case doesn't happen in updateUpperBoundofX() as we explicitly use the function pow(int n) with the bounds of Y. And since we loop on updating the bounds of variables X and Y, the case doesn't happen for X.UB, which is well updated wrt Y's bounds.

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropPowEven.java
ArthurGodet: I don't know, I did not touch the code wrt floor_nroot and ceil_nroot, but maybe.

jgFages: in my understanding, once double are involved, we should always use PRECISION to avoid numerical issues. 
@cprudhom any idea why it is used for PropPowOdd and not PropPowEven ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropPowEven.java
ArthurGodet: I just tested it and it seems to be the case. That said, given our use-case, I don't think it really matters (especially to maintain a propagator just for this). But I can add an if on the exponent, and call Math.sqrt whenever exponent = 2, if you want.

FYI, to compute all square root of integers from 0 to 100 000 000, it takes ~1100 ms on my computer with this formula, and ~1 ms with Math.sqrt.

jgFages: interesting! I am a bit surprised by the 1ms to compute 100 000 000 things but anyway I would normally suggest to add an IF to catch the case exp = 2. 

However, let's wait the answer of @cprudhom regarding the PRECISION. Because if precision is required, then we should probably stick to the pow

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropPowEven.java
ArthurGodet: It is interesting to note that the result of Math.pow has a precision of 1 ulp (https://docs.oracle.com/javase/6/docs/api/java/lang/Math.html#pow%28double,%20double%29), which is about 4e-16 in Java (https://en.wikipedia.org/wiki/Unit_in_the_last_place#Example_2).

If we are afraid of cases like 2.999999999 as a result instead of 3, we could always add 1e-9 for UB (and remove 1e-9 for LB). It should be OK then, do you agree ?

jgFages: I feel it would be safer yes

jgFages (approved):

Issue #1062, opened 2023-10-03
[BUG] Choco does not check for timeout during initialization phase


mstrojny:
**Describe the bug**
The timeout has no effect with specific conditions, which make loop in PropagationEngine.propagate() run long. The Solver is stuck in initialize() method.

**To Reproduce**
The example will have no solutions:
```
import org.chocosolver.solver.Model;
import org.chocosolver.solver.Solver;
import org.chocosolver.solver.constraints.Constraint;
import org.chocosolver.solver.variables.IntVar;

public class ChocoTimeout
{
    @SuppressWarnings("nls")
    public static void main(String [] args)
    {
        Model model = new Model();
        
        IntVar v7 = model.intVar("@v7", IntVar.MIN_INT_BOUND, IntVar.MAX_INT_BOUND, true);
        IntVar v7mi2 = model.sum("v7-2", v7, model.intVar(-2));
        
        Constraint arithm1 = model.arithm(v7, ">", v7);
        Constraint arithm2 = model.arithm(v7, "<", v7);
        Constraint arithm3 = model.arithm(v7, ">=", v7mi2);
        Constraint arithm4 = model.arithm(v7, "!=", model.intVar(5));
        model.post(arithm1, arithm2, arithm3, arithm4);
        
        Solver solver = model.getSolver();
        solver.limitTime(250);
        
        long start = System.currentTimeMillis();
        boolean solved = solver.solve();
        long took = System.currentTimeMillis() - start;
        
        System.out.println("Solved: " + solved);
        System.out.println("Time in measures: " + (solver.getMeasures().getTimeCount() * 1000));
        System.out.println("Time in solve(): " + took);
    }
}
```
the loop runs ~3 seconds on my machine with time limitation of 250 millis:
```
Solved: false
Time in measures: 3741.5261
Time in solve(): 3743
```
For more complicated models, it would run up to eternity.

**Expected behavior**
Timeout works for all phases of solver.solve().

**Environment:**
- Choco-solver version: [4.10.13](https://github.com/chocoteam/choco-solver/releases/tag/v4.10.13)

cprudhom:
The minimal _working_ example is:
```java
Model model = new Model();

IntVar v7 = model.intVar("@v7", IntVar.MIN_INT_BOUND, IntVar.MAX_INT_BOUND, true);

Constraint arithm1 = model.arithm(v7, ">", v7);
Constraint arithm2 = model.arithm(v7, "<", v7);
[...]

PR #1064, opened 2023-10-09
Make propagation stops sooner if time limit is met (fix issue #1062)

ArthurGodet:


solver/src/main/java/org/chocosolver/solver/propagation/PropagationEngine.java
```diff
@@ -189,6 +189,9 @@ public void propagate() throws ContradictionException {
             manageModifications();
             for (int i = nextNotEmpty(); i > -1; i = nextNotEmpty()) {
                 assert !pro_queue[i].isEmpty() : "try to pop a propagator from an empty queue";
+                if (model.getSolver().isTimeLimitMet()) {
```
jgFages, line 192: why not calling model.getSolver().isStopCriterionMet() ? 
This would avoid modifying the Solver class.

solver/src/main/java/org/chocosolver/solver/propagation/PropagationEngine.java
```diff
@@ -189,6 +189,9 @@ public void propagate() throws ContradictionException {
...
+                    return;
```
jgFages, line 193: Stopping the propagation there might lead to thinking the current state is valid / we found a solution. 
Therefore, I would rather throw an exception to make sure the current state is invalid, with something like this 

public final static ICause PROPAG_TIMEOUT = new ICause() {};
model.getSolver().throwsException(PROPAG_TIMEOUT, null, message);

Even if we haven't proved the state is infeasible either...

solver/src/test/java/org/chocosolver/solver/LimitsTest.java
```diff
@@ -35,6 +38,35 @@ public void testTime() {
...
+        // TODO : this test might need to be changed if better model analysis is done during model declaration
+
+        Solver solver = model.getSolver();
+        solver.limitTime(250);
+
+        long start = System.currentTimeMillis();
+        boolean solved = solver.solve();
+        long took = System.currentTimeMillis() - start;
+
+        assertFalse(solved);
```
jgFages, line 61: Good to check that solve() returns false;

solver/src/main/java/org/chocosolver/solver/propagation/PropagationEngine.java
ArthurGodet: The idea was to have a simple check on the time limit, and not check all the stopping criterion before the propagation of each propagator.
Indeed, all stopping criterion but the time limit do not depend on the problem's state outside of fixpoint. Unless you can think of one that would.

cprudhom: Personnaly, I would have move the code to the end of the while-loop just to keep the main purpose (ie, propagating) clear

cprudhom (approved):

Issue #564, opened 2017-10-31
Improvements of IntVar Views


arnaud-m:
Let x be a integer variable. The following views are available:

- scale view: s(x) = a*x  (a >= -1)
- minus view:  m(x) = -x
- offset view: t(x) = x + b

Here are two natural improvements
- [ ] extend the scale view for any a.
- [ ]  propose an affine view: f(x) = a*x + b

Experienced with choco-solver-_*{4.0.6}*_

PR #1065, opened 2023-10-11
Create of `IntAffineView`

cprudhom:
Related to #564 

This PR aims at merging the 3 views on `IntVar`, namely `IntOffsetView`, `IntScaleView` and `IntMinusView`, into a single one, namely `IntAffineView`.
Benefits are: 
- one class instead of 3;
- merging is easier: if `y = a.x + b` then `z = dy + c` is declared as `z = (d.a).x + d.b + c`. This limits the depth of expressions and moves events closer to the true variable;
- management of a in Z (instead of N in the current version of scale, which requires to declare a minus view first when a < 0);
- only one API for view: `model.intView(x, a, b)`.
Loss are:
- only one API for view: `model.intView(x, a, b)`, for instance, creating an offset view requires to declare two parameters (`a=-1` and `b=0`) instead of 0 in the previous version.

In addition, it seems that the recognition of views were incorrect, on some problems, it reduces the number of views (mostly when `Task` are in the game).
For instance, `MSPSP-easy_01.xml.lzma` (xcsp) previously created 370 including 149 views where now 241 variables are created including 20 views.
Inevitably, it pertubs black-box strategies and some deterioration can be observed.

Note: most of the modifications in this PR are related to code adaptation.
The main classes to double check are `IntAffineView` (+ tests) and `IViewFactory`. 

**Note**: _I also define default behavior in `IntVar` for some advanced methods (like removeInterval, for example). This is related to this PR since the affine view only defines the 4 main methods._


solver/src/main/java/org/chocosolver/solver/variables/IViewFactory.java
```diff
@@ -109,138 +108,84 @@ default BoolVar[] setBoolsView(SetVar setVar, int size, int offset) {
...
-     * @param cste a constant (can be either negative or positive)
-     * @return an IntVar equal to <i>var+cste</i>
+     * Create an affine view based on <i>var</i>, i.e. <i>view = a * var + b</i>
+     *
+     * @param var an integer variable
+     * @param a   a coefficient
+     * @param b   a constant
+     * @return an affine view
      */
-    default IntVar intOffsetView(IntVar var, int cste) {
```
jgFages, line 117: I would prefer to keep the existing methods and replace their content by a redirection to the new intview method, in order to ensure backward compatibility (which would have avoided the rewriting of the tests).

cprudhom: Personnally, I found the old names too long/verbose and it's time for a change :)
Note that the method are deprecated but still there.
An alternative could be to have shorter name, like: `offset(IntVar, int)`, `scale(IntVar, int)` and `minus(IntVar)`. (👍 / 👎 )

solver/src/main/java/org/chocosolver/solver/variables/IViewFactory.java
```diff
@@ -109,138 +108,84 @@ default BoolVar[] setBoolsView(SetVar setVar, int size, int offset) {
...
-     * Creates a view over <i>var</i> equal to -<i>var</i>.
-     * That is if <i>var</i> = [a,b], then this = [-b,-a].
-     *
-     * @param var an integer variable
-     * @return an IntVar equal to <i>-var</i>
+     * @see #intView(IntVar, int, int)
+     * @deprecated
+     */
+    @Deprecated
+    default IntVar intOffsetView(IntVar var, int cste) {
```
jgFages, line 169: ah ok it is still there.

solver/src/main/java/org/chocosolver/solver/variables/IViewFactory.java
jgFages: Thanks! My main concern was the backward compatibility for other models. If they are still there it is fine. 

Regarding alternatives (in addition to historical names), I would suggest to harmonize with expressions, and therefore use :  
- add
- mul
- minus (minus does not already exists in the expressions but I think it would be very usefull to add it).

cprudhom: We need to be sure that there is no ambiguity between (`x` is a variable ,`a` and `b` ints and `m` a model):
- `x.neg()` (which is equivalent to `minus` view), `x.mul(a)`, `x.add(b)` and
- `m.neg(x)`, `m.mul(x, a)` and `m.add(x, b)`.
Following this, we can add `m.aff(x, a, b)` (or `m.aff(a, x, b)` and probably need to detect expressions like : `x.mul(a).add(b)` to a view (which is not the easy part).

jgFages: I didn't have the neg() in mind thanks. Even if I understand your point about symetry, I would prefer m.minus(x) than m.neg(x) because it is closer to the natural language, but maybe m.neg(x) is ok too. 

Also, I sometimes wish we could have x.minus(y) to replace x.add(y.neg())... 

Then you opened two discussions

1. the naming **aff** versus **intView** => I prefer **intView**as you put in the PR, for the following reasons
- I believe that if a user uses the view instead of x.mul(a).add(b), it is because he is aware that performances my be better using views, therefore the name intView seems relevant. 
- Whereas I may say "y is cosinus x", I never say "y is an affine function of x, a, b..." so having such names in the API is not intuitive to me. 
2. the argument order => I prefer the following : a, x, b, closer to the mathematical writing and emphasing on the difference between a and b.

cprudhom: I anticipated too much, I'm going to retro-pedal. 
May be we can wait for @ArthurGodet opinion before

cprudhom: But, the ideal solution is to manage views through expressions, it is even more natural.
I'll think of a way to do that.
On other thing is: do we really need to keep `Settings#enableViews` which is (I suppose) never changed. 
_The discussion might concern other settings, but it is out of this PR._

solver/src/main/java/org/chocosolver/solver/variables/view/integer/IntAffineView.java
```diff
@@ -0,0 +1,588 @@
...
+    }
+
+    @Override
+    public int nextValue(int v) {
+        if (v < getLB()) return getLB();
+        if (v > getUB()) return Integer.MAX_VALUE;
+        // y = p * a * x + b where p in {-1,1}, a and b in Z.
+        // => x = (y - b) / (p * a)
+        v -= b;
+        v = (v < 0 && a > 1) ? v - 1 : v;
```
ArthurGodet, line 440: Should be replaced by `v = (v < 0 && a > 1 && v % a != 0) ? v - a : v;`

See below why

solver/src/main/java/org/chocosolver/solver/variables/view/integer/IntAffineView.java
```diff
@@ -0,0 +1,588 @@
...
+        v /= (p ? a : -a);
+
+        if (p) {
+            v = var.previousValue(v);
+        } else {
+            v = var.nextValue(v);
+        }
+        if (v == Integer.MIN_VALUE || v == Integer.MAX_VALUE) return Integer.MIN_VALUE;
+        return (p ? v : -v) * a + b;
+    }
```
ArthurGodet, line 494: This function is not correct. I tested it with the following code : 
`Model model = new Model();
 IntVar x = model.intVar(2, 9);
 IntVar y = model.intView(x, 5, 7);`

And calling `previousValue` has a "shifted" behaviour. For instance, `previousValue(51)` and `previousValue(52)` return 47 as they should, but `previousValue(48)`, `previousValue(49)` and `previousValue(50)` return 42 instead of 47 (and this "shift" is repeated when looping down the values).

This unwanted behaviour is due to the division done in line 485. Indeed, line 484 is not enough to compensate for it. Replacing line 484 by `v = (v > 0 && a > 1 && v % a != 0) ? v + a : v;` fixes it.


I think we should add a test checking this behaviour, assuring that from a value `v` such that `v % a == 0`, the `a` values above (from `v` to `v + (a - 1)`) all return `v + a` as `nextValue`, and the `a` values below (from `v - a + 1` to `v`) all return `v - a` as `previousValue`.

solver/src/main/java/org/chocosolver/solver/variables/view/integer/IntAffineView.java
```diff
@@ -0,0 +1,588 @@
...
+
+    @Override
+    public boolean hasEnumeratedDomain() {
+        return var.hasEnumeratedDomain() || a > 1;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof IntAffineView)) return false;
```
ArthurGodet, line 550: I make this remark here, but it stands for the entire class : I am not a fan of inline if, for, while, and so on (but some standard formats allow them for short lines such as these ones). I know that we did not choose formatting rules for the code in Choco, so it is just a remark, you might leave it this way.

ArthurGodet (changes requested):

solver/src/main/java/org/chocosolver/solver/variables/view/integer/IntAffineView.java
```diff
@@ -0,0 +1,588 @@
...
+        boolean change;
+        if (!p) {
+            change = var.updateUpperBound(-value - 1, this);
+        } else {
+            change = var.updateLowerBound(value + 1, this);
+        }
+        IntEventType e = IntEventType.INCLOW;
+        if (isInstantiated()) {
+            e = IntEventType.INSTANTIATE;
+        }
```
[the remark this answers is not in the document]
cprudhom: If `change` is false, the event will be ignored, otherwise, it means that the modification led to instantiation.

solver/src/main/java/org/chocosolver/solver/variables/view/integer/IntAffineView.java
```diff
@@ -0,0 +1,588 @@
...
+        boolean change = false;
+        if (!p) {
+            change = var.updateLowerBound(-value, this);
+        } else {
+            change = var.updateUpperBound(value, this);
+        }
+        IntEventType e = IntEventType.DECUPP;
+        if (isInstantiated()) {
+            e = IntEventType.INSTANTIATE;
+        }
```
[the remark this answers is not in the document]
cprudhom: Same answer as above

solver/src/main/java/org/chocosolver/solver/variables/view/integer/IntAffineView.java
```diff
@@ -0,0 +1,588 @@
...
+    }
+
+
+    @Override
+    public boolean removeValue(int value, ICause cause) throws ContradictionException {
+        assert cause != null;
+        int inf = getLB();
+        int sup = getUB();
+        if (inf > value || value > sup) return false;
+        model.getSolver().getEventObserver().removeValue(this, value, cause);
```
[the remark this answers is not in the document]
cprudhom: This is required for explanation purpose only, the explanation of var's modification is because the view has been modified in the first place, so the implication graph must reflect this order.

solver/src/main/java/org/chocosolver/solver/variables/IViewFactory.java
cprudhom: ### neg or minus
The name takes its origin in the XCSP3 documentation ([the table of operators](http://xcsp.org/specifications/constraints/generic/intension/)).
So I suppose, we can keep `neg()` as the final choice.

### add `view` to API
I can live with that. I'll change this in the next commit.
The methods will be: 
- `intView(int a, IntVar var, int b)` : the default one (ie, affine)
- `absView(IntVar var)`
- `addView(IntVar var, int b)` : I would have preferred `offsetView` but it is not close to the expression DSL and it is quite _long_
- `mulView(IntVar var, int a)`: here again, I would have preferred `scaleView`, but OK
- `negView(IntVar var)`

### Parameters ordering
I choose: `int a, IntVar var, int b`.

### Settings#enableViews
I will integrate this in the current PR, unless someone is really against it.

### manage views through expressions
This should be done in another PR

cprudhom: Thinking twice, `addView ` might be ambiguous: one can think it adds something to the model.
What do you think? `plusView`, `offsetView` or `addView` ?

cprudhom: `Settings#enableViews` => that's very helpful for testing, I'll keep it 😀

ArthurGodet: Maybe we did not understand each other. I was for keeping it like you did : add(), neg(), mul() without the view in the name. Indeed, we should say that they return views, and have the word view only in the intView() function.

But I agree that add() might be ambiguous, so maybe just offset() ?

solver/src/main/java/org/chocosolver/solver/variables/view/integer/IntAffineView.java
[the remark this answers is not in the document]
ArthurGodet: Yes, it is just that the IntEventType needs to be declared only if change is true.

cprudhom:
Are there still remarks? Or can one of you approuve the PR?

ArthurGodet (approved):

PR #1066, opened 2023-10-20
Remove randomness by default of BBox strategies

cprudhom:
The main idea is to define how ties are broken in black-box strategies.
Previously, ties were broken randomly, which is a kind of admission of powerlessness.
I propose, with this PR,  to make possible to declare a tie-breaker if needed or let the default one which is based on  lexical ordering (=variables input order).
Note that this change has few mainly an impact on fresh start when no score are recorded, but as it drives the search to different search space, it may have a huge impact (+ or -) on efficiency.
However, now, we can tweak it.


ArthurGodet (approved):

cprudhom:
I think we should revert the PR and wait for the checks to pass.
@ArthurGodet What do you think?

PR #1067, opened 2023-10-22
fix the creation of IntAffineView based on another IntAffineView

ArthurGodet:


solver/src/main/java/org/chocosolver/solver/variables/IViewFactory.java
```diff
@@ -134,7 +134,7 @@ default IntVar intView(int a, IntVar var, int b) {
             if (var instanceof IntAffineView) {
                 IntAffineView<?> view = (IntAffineView<?>) var;
                 int av = (view.p ? 1 : -1) * view.a * a;
```
jgFages, line 136: General question : I was wondering what p stood for and I understood that once the view is created, a becomes |a|. 
Is it historical legacy or is the code simpler assuming a is positive ? 
I see several `(p ? 1 : -1) * a` and therefore think the code could be greatly simplified by keeping a as it is (i.e. either positive or negative).

Well done for the bug fix.

jgFages (dismissed):

solver/src/main/java/org/chocosolver/solver/variables/IViewFactory.java
cprudhom: It was clearer for me to consider `a` as positive, that's all :)

cprudhom (dismissed):

cprudhom (approved):

PR #1076, opened 2024-01-09
Fix PropCount_AC wrt bounded domains

ArthurGodet:


solver/src/main/java/org/chocosolver/solver/constraints/nary/count/PropCount_AC.java
```diff
@@ -112,15 +112,17 @@ public void propagate(int varIdx, int mask) throws ContradictionException {
...
                 if (!vars[varIdx].contains(value)) {
                     possibles.remove(varIdx);
-                    filter();
+                    forcePropagate(PropagatorEventType.CUSTOM_PROPAGATION);
                 } else if (vars[varIdx].isInstantiated()) {
                     possibles.remove(varIdx);
                     mandatories.add(varIdx);
-                    filter();
+                    forcePropagate(PropagatorEventType.CUSTOM_PROPAGATION);
+                } else if (!vars[varIdx].hasEnumeratedDomain() && (vars[varIdx].getLB() == value || vars[varIdx].getLB() == value)) {
```
jgFages, line 120: repeated condition, the second should be getUB() ?

ArthurGodet: Indeed, the second condition should be on getUB()

jgFages: I can see that calling filter() instead of forcePropagate can lead to interferences with the incremental approach (although if you have a detailed explanation, I am interested).

But why is there a specific condition on bounded domain ? I do not see what was missing.

Thanks

ArthurGodet: I use forcePropagate() instead of directly filter() only for performance (to avoid calling several time filter() where one call is enough).

For the condition on bounded domain, it is to force a propagation that was impossible to enforce before : let say that the counting variable is instantiated and mandatories has the right number of elements. In such a case, propagation would remove from value from the domain of variables in possibles (and remove the variables from possibles if the removal of value was done). However, for variables with bounded domains, it might be impossible to remove the value (if it is not a bound of the domain). Therefore, the removal should happen as soon as it is possible.

Now that I think about it, if the lower bound of a bound-domain variable (the reasoning is similar for upper bound) is done such that var.LB > value, then the propagator whould not be set passive and the variable removed from possibles. I can't think of a case where that would impact the solution (propagator is correct in terms of propagation) but updates in variables' bounds would still call the propagate(int idxVarInProp, int mask). I guess that the conditions `(vars[varIdx].getLB() >= value || vars[varIdx].getUB() <= value))` would set the propagator to passive correctly.

jgFages: yes I think (vars[varIdx].getLB() >= value || vars[varIdx].getUB() <= value)) should be better

cprudhom (approved):

solver/src/main/java/org/chocosolver/solver/constraints/nary/count/PropCount_AC.java
ArthurGodet: Actually, using the conditions with `==` is correct as otherwise the condition `!vars[varIdx].contains(value)` whould be true and the first if will force the propagation.

Issue #1112, opened 2024-11-08
[BUG] BinPacking initial propagation does not detect inconsistency


cprudhom:
The following model shows how to reproduce the bug: item1 of size 9 is assigned to bin 0 with capacity 0.
This inconistency should be detected on initial propagation but the initialisation of `R` and `sumR`

The test:
```java
@Test(groups = "1s", timeOut = 60000)
    public void testBug1() {
        Model model = new Model();
        int[] itemSize = new int[]{2, 9};//, 6, 1, 0, 5};
        IntVar[] itemBin = new IntVar[2];
        itemBin[0] = new BitsetIntVarImpl("X_0", new int[]{0, 1}, model);
        itemBin[1] = new BitsetIntVarImpl("X_1", new int[]{0}, model);
        IntVar[] binLoad = new IntVar[2];
        binLoad[0] = new BoolVarImpl("X_6", model);
        binLoad[1] = new BitsetIntVarImpl("X_8", new int[]{0, 9, 11}, model);
        Constraint c = model.binPacking(itemBin, itemSize, binLoad, 0);
        c.post();
        System.out.println(c.isSatisfied()); // print FALSE
        Solver solver = model.getSolver();
        try {
            solver.propagate();
            System.out.println(c.isSatisfied()); // print FALSE
            Assert.fail(); // then assertion fails...
        }catch (ContradictionException e) {
            // should fail
        }
//        System.out.println(c.isSatisfied());
    }
```

cprudhom:
A fix is to add the following filtering rules during initial propagation:

```java
// ...
for(int i = 0; i < binLoad.length; i++) { // Load All
    binLoad[i].updateLowerBound(sumR[i].get(), this);
    binLoad[i].updateUpperBound(sumP[i].get(), this);
}
binsToProcess.set(0, nbAvailableBins);
//...
```

ArthurGodet:
I guess the fix is to delete the following lines from the constructor :
`if(itemBin[i].isInstantiated()) {
                        R[j].add(i);
                        rj += itemSize[i];
                    }`

Indeed, the problem comes from the fact that elements already in R (because of these lines in the constructor) are not well updated in method `updateRAfterInstantiation`

PR #1113, opened 2024-11-09
Fix bin packing initial propagation

ArthurGodet:
Fix bin packing's propagator initial propagation (described in #1112)

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/constraints/nary/binPacking/PropBinPacking.java
+++ b/solver/src/main/java/org/chocosolver/solver/constraints/nary/binPacking/PropBinPacking.java
@@ -127,16 +127,8 @@ public class PropBinPacking extends Propagator<IntVar> {
         for(int j = 0; j<nbAvailableBins; j++) {
             P[j] = SetFactory.makeStoredSet(SetType.BITSET, 0, itemBin[0].getModel());
             R[j] = SetFactory.makeStoredSet(SetType.BITSET, 0, itemBin[0].getModel());
-            int pj = 0;
-            int rj = 0;
-            for(int i = 0; i<nbItems; i++) {
-                if(itemBin[i].contains(j+offset)) {
-                    P[j].add(i);
-                    pj += itemSize[i];
-                }
-            }
-            sumR[j] = itemBin[0].getModel().getEnvironment().makeInt(rj);
-            sumP[j] = itemBin[0].getModel().getEnvironment().makeInt(pj);
+            sumR[j] = itemBin[0].getModel().getEnvironment().makeInt(0);
+            sumP[j] = itemBin[0].getModel().getEnvironment().makeInt(0);
         }
 
         binsToProcess = new BitSet(nbAvailableBins);
@@ -344,6 +336,16 @@ public class PropBinPacking extends Propagator<IntVar> {
     @Override
     public void propagate(int evtmask) throws ContradictionException {
         if(PropagatorEventType.isFullPropagation(evtmask)) {
+            for(int j = 0; j<nbAvailableBins; j++) {
+                int pj = 0;
+                for (int i = 0; i < nbItems; i++) {
+                    if (itemBin[i].contains(j + offset)) {
+                        P[j].add(i);
+                        pj += itemSize[i];
+                    }
+                }
+                sumP[j].set(pj);
+            }
             for(int i = 0; i<itemBin.length; i++) { // Pack All
                 itemBin[i].updateBounds(offset, nbAvailableBins+offset-1, this);
                 if(itemBin[i].isInstantiated()) {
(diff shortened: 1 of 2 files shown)


solver/src/main/java/org/chocosolver/solver/constraints/nary/binPacking/PropBinPacking.java
```diff
@@ -133,10 +133,6 @@ public PropBinPacking(IntVar[] itemBin, int[] itemSize, IntVar[] binLoad, int of
                 if(itemBin[i].contains(j+offset)) {
```
jgFages, line 133: shouldn't this whole block (setting of P and R) be inside the fullPropagation method (last method) ?

ArthurGodet: It could, and it might be more consistent in terms of code. In terms of correctness, I think it does not change anything so  either way (let the code inside the constructor or at the beginning of the fullPropagation part) is ok for me

jgFages: Yes, it is mostly for readability. 
I remember though having encountered an issue with @cprudhom many years ago because of setting up data structure in the constructor, in a specific context where a fullPropagation was called during search by the engine. Unfortunately, I do not remember the details and maybe the involved sets where not stored, in which case we understand easily why it didn't work at that time. @cprudhom does it ring a bell ?

ArthurGodet: I move the code, initialising P and sumP, from the constructor to the beginning of the fullPropagation.

solver/src/main/java/org/chocosolver/solver/constraints/nary/binPacking/PropBinPacking.java
```diff
@@ -348,6 +336,16 @@ public void propagate(int idxVarInProp, int mask) throws ContradictionException
     @Override
     public void propagate(int evtmask) throws ContradictionException {
         if(PropagatorEventType.isFullPropagation(evtmask)) {
+            for(int j = 0; j<nbAvailableBins; j++) {
```
jgFages, line 339: to be 100% safe, P[j] and R[j] should be cleared here I think

solver/src/main/java/org/chocosolver/solver/constraints/nary/binPacking/PropBinPacking.java
```diff
@@ -348,6 +336,16 @@ public void propagate(int idxVarInProp, int mask) throws ContradictionException
...
+                int pj = 0;
+                for (int i = 0; i < nbItems; i++) {
+                    if (itemBin[i].contains(j + offset)) {
+                        P[j].add(i);
+                        pj += itemSize[i];
+                    }
+                }
+                sumP[j].set(pj);
```
jgFages, line 347: now we are within the propagate method and no longer in the constructor, maybe we can reuse the 
```
if(itemBin[i].isInstantiated()) {
                        R[j].add(i);
                        rj += itemSize[i];
                    } 
                    sumP[j].set(rj);
```
associated with the filtering that was missing, if you think that doing so would make the code clearer.
If you prefer keeping it as it is, I am fine with that too

ArthurGodet: No need to, it will be done correctly at line 353

solver/src/main/java/org/chocosolver/solver/constraints/nary/binPacking/PropBinPacking.java
cprudhom: I agree, reifying this constraint could lead to unexpected results if `P[j]` and `R[j]` are not cleared.

cprudhom (changes requested):
I agree with the correction, but it seems safer not to clean up P and R in full propagation.

solver/src/main/java/org/chocosolver/solver/constraints/nary/binPacking/PropBinPacking.java
cprudhom: > @cprudhom does it ring a bell ?

As soon as there are data structures, you have to be careful about initializing them, that's true. As the variables may have been modified since the call to the constructor, the structures may not be up to date, which is why it is always better to delay populating them until the initial propagation.

ArthurGodet:
You mean, it would be safer to clean R and P in full propagation ? Or not clean them (as it is now) ?

cprudhom:
> You mean, it would be safer to clean R and P in full propagation ? Or not clean them (as it is now) ?

To set them to 0 as the first instruction of the `for-loop` (l.339)

solver/src/main/java/org/chocosolver/solver/constraints/nary/binPacking/PropBinPacking.java
ArthurGodet: Yes, I agree it would be safer (and it is now done)

cprudhom (approved):

jgFages (approved):

Issue #1114, opened 2024-11-15
[BUG]Unexpected behavior when using Task with scalar


cprudhom:
The following test fails:

```java
@Test(groups = "1s", timeOut = 60000)
public void testScalarAndTaskBug1() {
    int[] coeffs = new int[]{3, -2,-1,-2};
    IntVar[] vars = new IntVar[4];
    vars[0] = model.intVar(8, 9);
    vars[1] = model.intVar(2, 4);
    vars[2] = model.intVar(new int[]{1,2,4,5,6});
    vars[3] = model.intVar(3, 5);
    model.scalar(vars, coeffs, "<=", 1).post();
    IntVar ee = model.intVar(4, 9);
    new Task(vars[3], vars[2], ee);
    model.getSolver().solve();
    Assert.assertEquals(model.getSolver().getSolutionCount(), 0);
}
```
When the `scalar` constraint filters, it first reduces `vars[0]` to `8`, then `vars[1]=4` and `vars[2]={5,6}`.
At this point the task automatically reduces the domain of `vars[3]` to `{3,4}`.
Then the `scalar` constraint keeps on filtering and fix vars[3]` to `4` and passivates itself.
But, with `vars[3]=4` there is no solution anymore to this model and `scalar` is not able to detect it because it does not react on modification during its filtering step.
The combination of task filtering and passivate leds to inconsistency. 

So IMO, the way tasks (or more generally monitors) reduce domains is buggy, because we cannot consider that constraints should be aware of reductions made by monitors during their filtering step.
I'm afraid other constraints will produce the same bug...

jgFages:
It would be interesting to run a benchmark comparing monitors vs classical arithm. I wonder if the monitors are worth it. I am not sure the gain is significant (there are cases of performance loss), and the hack may have side effect on search and learning as well...

cprudhom:
There is (at least) one case where monitors are mandatory, it is when the `cumulative` constraint is declared.

jgFages:
It was the motivation long ago but is it really still necessary? For which algorithm?

ArthurGodet:
Yes, filtering done by monitors (and thus by TaskMonitor) can lead to very buggy behaviours inside propagators, as not all of them have been designed to adapt to such filtering. In the case of PropScalar, storing lb and ub values for each variable in the prepare() method would lead to correct behaviour (btw, it would be more consistent as the algorithm rely on pre-computed values such as the `I` array).

[...]

PR #1117, opened 2024-11-17
Fix faulty behaviour of #1114

ArthurGodet:
Replace the TaskMonitor by arithm constraints
Update code of Cumulative implementations to update `start + duration = end` relation as soon as possible

The only problem of this implementation is that it might slow down a bit the solving process, as arithm constraints and cumulative constraints might do a ping-pong (however, it should not happen if the relation is well propagated inside cumulative filter implementations, but I could not find where it was not done fine). I have spotted such a slow down in CumulativeTest.test6() which is now taking 1.5s, instead of 0.5s (but no slow down on other testing methods).

That said, I tested on the new implementations of cumulative and disjunctive constraints I am working on, and no slow down has been spotted in these. So it is up to you to wait for this bigger Pull Request, or to accept this quick fix.

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/CumulFilter.java
+++ b/solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/CumulFilter.java
@@ -56,4 +56,15 @@ public abstract class CumulFilter {
 	 * @throws ContradictionException
 	 */
 	public abstract void filter(IntVar[] s, IntVar[] d, IntVar[] e, IntVar[] h, IntVar capa, ISet tasks, Propagator<IntVar> aCause) throws ContradictionException;
+
+	protected void propStartDurationEndRelation(IntVar[] s, IntVar[] d, IntVar[] e, IntVar[] h, Propagator<IntVar> aCause) throws ContradictionException {
+		int n = s.length;
+		for (int i = 0; i < n; i++) {
+			d[i].updateLowerBound(0, aCause);
+			h[i].updateLowerBound(0, aCause);
+			s[i].updateBounds(e[i].getLB() - d[i].getUB(), e[i].getUB() - d[i].getLB(), aCause);
+			e[i].updateBounds(s[i].getLB() + d[i].getLB(), s[i].getUB() + d[i].getUB(), aCause);
+			d[i].updateBounds(e[i].getLB() - s[i].getUB(), e[i].getUB() - s[i].getLB(), aCause);
+		}
+	}
 }
--- a/solver/src/main/java/org/chocosolver/solver/variables/Task.java
+++ b/solver/src/main/java/org/chocosolver/solver/variables/Task.java
@@ -9,15 +9,14 @@
  */
 package org.chocosolver.solver.variables;
 
+import org.chocosolver.solver.ICause;
 import org.chocosolver.solver.Model;
+import org.chocosolver.solver.constraints.Constraint;
 import org.chocosolver.solver.exception.ContradictionException;
 import org.chocosolver.solver.learn.ExplanationForSignedClause;
-import org.chocosolver.solver.variables.events.IEventType;
-import org.chocosolver.solver.variables.events.IntEventType;
 import org.chocosolver.solver.variables.view.integer.IntAffineView;
 import org.chocosolver.util.objects.setDataStructures.iterable.IntIterableRangeSet;
 
-import java.util.ArrayList;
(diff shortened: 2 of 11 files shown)


ArthurGodet:
Failing tests seem to come from performance issues of the Cumulative constraint (for xcsp failing test, it seems to come from a greater number of nodes, which is a bit odd, as new implementation should not induce a weaker filtering).

To investigate

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/CumulFilter.java
```diff
@@ -56,4 +56,15 @@ public CumulFilter(int nbMaxTasks){
...
 	public abstract void filter(IntVar[] s, IntVar[] d, IntVar[] e, IntVar[] h, IntVar capa, ISet tasks, Propagator<IntVar> aCause) throws ContradictionException;
+
+	protected void propStartDurationEndRelation(IntVar[] s, IntVar[] d, IntVar[] e, IntVar[] h, Propagator<IntVar> aCause) throws ContradictionException {
+		int n = s.length;
+		for (int i = 0; i < n; i++) {
+			d[i].updateLowerBound(0, aCause);
+			h[i].updateLowerBound(0, aCause);
+			s[i].updateBounds(e[i].getLB() - d[i].getUB(), e[i].getUB() - d[i].getLB(), aCause);
+			e[i].updateBounds(s[i].getLB() + d[i].getLB(), s[i].getUB() + d[i].getUB(), aCause);
+			d[i].updateBounds(e[i].getLB() - s[i].getUB(), e[i].getUB() - s[i].getLB(), aCause);
```
jgFages, line 67: if domains are enumerated, you might need to loop

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/CumulFilter.java
```diff
@@ -56,4 +56,15 @@ public CumulFilter(int nbMaxTasks){
 	 * @throws ContradictionException
 	 */
 	public abstract void filter(IntVar[] s, IntVar[] d, IntVar[] e, IntVar[] h, IntVar capa, ISet tasks, Propagator<IntVar> aCause) throws ContradictionException;
+
+	protected void propStartDurationEndRelation(IntVar[] s, IntVar[] d, IntVar[] e, IntVar[] h, Propagator<IntVar> aCause) throws ContradictionException {
```
jgFages, line 60: what if we don't do that at all and leave it to arithm ?

ArthurGodet: As I said, the idea is to avoid a ping-pong between arithm constraints and the cumulative one, as the latter could be long to initiate.  But maybe it is OK to allow it

ArthurGodet: Ok I found a new reason : some filtering algorithms rely on the hypothesis of the task relation, especially the one in NRJCumulFilter

But I still can't really explain why we have the performance issues

cprudhom: One possible way to avoid iterating over all tasks is to use IVariableMonitor to store modified tasks and only iterate over them. This is, for example, what is done in ActivityBased search.

solver/src/main/java/org/chocosolver/solver/variables/Task.java
```diff
@@ -29,7 +28,7 @@
  * @author Jean-Guillaume Fages
  * @since 04/02/2013
  */
-public class Task {
+public class Task implements ICause {
```
cprudhom, line 31: I'm confused: the Task object was initially designed to maintain the relationship between start, duration and end during propagation. Now, if the relation is maintained explicitly by propagators, what is the purpose of Task ?

solver/src/main/java/org/chocosolver/solver/variables/Task.java
```diff
@@ -147,7 +151,15 @@ private void declareMonitor() {
      * @throws ContradictionException thrown if a inconsistency has been detected between start, end and duration
      */
     public void ensureBoundConsistency() throws ContradictionException {
```
cprudhom, line 153: Here again, is this method still relevant?

cprudhom (changes requested):

Issue #1141, opened 2025-02-21
`TIntArrayList.clear()` unneccesarily allocates massive amounts of memory thus degrading performance


jsimomaa:
Noticed this while profiling my application. Trove's `TIntArrayList.clear()` allocates lots of `int[]`.

![Image](https://github.com/user-attachments/assets/d6ebf161-680d-45bf-beac-becceb067698)

This can be fixed by replacing `clear()` with `reset()` (or event `resetQuick()`):

https://trove4j.sourceforge.net/javadocs/gnu/trove/list/array/TIntArrayList.html#reset()

cprudhom:
Fix by #1143

PR #1142, opened 2025-02-21
Change clear() to reset / resetQuick to prevent allocating unneccessa…

jsimomaa:
…ry int[] backing arrays #1141


solver/src/main/java/org/chocosolver/solver/constraints/nary/nvalue/PropAtMostNValues_BC.java
```diff
@@ -157,12 +157,12 @@ private boolean pruneLB() throws ContradictionException {
...
         boolean hasChanged = vars[n].updateLowerBound(nbKer, this);
         if (nbKer == vars[n].getUB()) {
-            stamp.clear();
+            stamp.reset();
             for (int i = 0; i < n; i++) {
                 node = orderedNodes[i];
                 if (kerRepresentant.get(node)) {
                     hasChanged |= updateKer(minVal[node], true);
-                    stamp.clear();
+                    stamp.reset();
```
jgFages, line 165: I think we could use resetQuick here as well

jgFages:
Hi, thank you very much for the improvement, I was not aware of that ! 
I feel it would be worth going through all usages of trove lists within the library.

cprudhom:
I just changed the targeted branch to `develop` instead of `master`

jgFages:
So it would need a rebase from develop right? I see 32 commits

cprudhom:
Hmm, it could be a better idea to create a new PR from from `develop`

jgFages:
I close this PR and will do another one going through all usages of clear() in the solver. Thanks again

PR #1143, opened 2025-02-21
replace .clear() by .resetQuick() for TIntArrayList

jgFages:


ArthurGodet (approved):

cprudhom:
It seems that something needs to be checked for LNS because the test fails in 919ms.
I'll have a look today.

PR #1146, opened 2025-02-25
Split extract MiniSatSolver from MiniSat

cprudhom:
Everything related to search in MiniSat is useless for LCG. So, a subclass can be created only for that purpose, namely MiniSatSolver.

+ activate topLevelCleanUp in LCG

### DIFF
--- a/solver/src/main/java/org/chocosolver/sat/MiniSat.java
+++ b/solver/src/main/java/org/chocosolver/sat/MiniSat.java
@@ -10,17 +10,13 @@
 package org.chocosolver.sat;
 
 import gnu.trove.list.TIntList;
-import gnu.trove.list.array.TDoubleArrayList;
 import gnu.trove.list.array.TIntArrayList;
 import gnu.trove.map.hash.TIntObjectHashMap;
 import org.chocosolver.solver.variables.impl.LitVar;
-import org.chocosolver.util.ESat;
-import org.chocosolver.util.objects.IntHeap;
 
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Comparator;
-import java.util.Random;
 
 /**
  * <p>A MiniSat solver.</p>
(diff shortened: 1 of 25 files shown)


solver/src/main/java/org/chocosolver/sat/MiniSat.java
```diff
@@ -922,35 +747,45 @@ private int getBacktrackLevel(TIntList out_learnt) {
...
+            if (satisfied(c)){
+                removeClause(cs.get(i));
+            }else{
+                cs.set(j++, cs.get(i));
+
+            }
+        }
+//        cs.shrink(i - j);
+        if (cs.size() > j) {
+            cs.subList(j, cs.size()).clear();
```
jgFages, line 775: rebase branch and check replacement by .resetQuick()

solver/src/main/java/org/chocosolver/sat/MiniSatSolver.java
```diff
@@ -0,0 +1,327 @@
...
+     * A call to this method will attempt to find
+     * an interpretation that satisfies the Boolean formula declared in this.
+     *
+     * @return {@code ESat.TRUE} if such an interpretation is found,
+     * {@code ESat.FALSE} if no interpretation exists,
+     * {@code ESat.UNDEFINED} if a limit was reached.
+     */
+    public ESat solve() {
+        model.clear();
+        conflict.clear();
```
jgFages: rebase branch and check replacement by .resetQuick()

solver/src/main/java/org/chocosolver/sat/MiniSatSolver.java
```diff
@@ -0,0 +1,327 @@
...
+
+        for (; ; ) {
+            propagate();
+            if (confl != C_Undef) {
+                // CONFLICT
+                conflicts++;
+                conflictC++;
+                if (trailMarker() == 0) return ESat.FALSE;
+
+                learnt_clause.clear();
```
jgFages: rebase branch and check replacement by .resetQuick()

solver/src/main/java/org/chocosolver/sat/MiniSat.java
cprudhom: here, `cs.subList` returns a view of cs as a `List` without the `resetQuick()` method.

PR #1152, opened 2025-03-25
Faster search in scope

jgFages:


solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/ConflictOrderingSearch.java
```diff
@@ -98,7 +87,7 @@ public ConflictOrderingSearch(Model model, AbstractStrategy<V> mainStrategy) {
 
     @Override
     public boolean init() {
-        if(model.getSolver().getSearchMonitors().contains(this)) {
+        if (!model.getSolver().getSearchMonitors().contains(this)) {
```
jgFages, line 90: Here was the bug @cprudhom

solver/src/main/java/org/chocosolver/solver/search/strategy/strategy/AbstractStrategy.java
```diff
@@ -9,91 +9,134 @@
...
+    public boolean isVarInScope(Variable var) {
+        return idScope.get(var.getId());
     }
 
     /**
-     * @return array of variables
+     * Prepare <code>this</code> to be used in a search loop
+     * The initialization can detect inconsistency, in that case, it returns false
      */
-    public V[] getVariables() {
```
cprudhom, line 84: Why removing this method? Is it not used elsewhere?

jgFages: The method is just moved. The display on github is poor, better to look at the branch locally

PR #1154, opened 2025-03-27
Performance improvement of criterion-based search strategies

cprudhom:
Include:
- [x] correction of ConflictHistorySearch and its simplification
- [x] introduction of `Weigthable` interface and two implementations (as an array and as a map)
- [x] `weights` of `AbstractCriterionBasedVariableSelector` are now stored in `Weigthable.Array`
- [x] simplification of `AbstractCriterionBasedVariableSelector`: some structures have been moved to DomOverWDeg for clarity

Could also include:
- [ ] Split `Model.nextId()` into `Model.nextVarId()` and `Model.nextPropId()`. Currently variables and propagators implement `Identity` interface and have though a unique id (`getId()`). These ids are generated thanks to `Model.nextId()` and ensure uniqueness of values between variables and propagators. I don't think this is necessary and could limit the size of structures that depend on `ID`s.
- [x] `PickOnDom` cannot be simply adapted to constaint-oriented propagation engine. Something must be done (remove it, adapt it, etc). 
- [ ] set the orientation of the propagation engine once and for all


solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/DomOverWDeg.java
```diff
@@ -99,15 +215,92 @@ final int remapInc() {
...
+    /// ///////////////////////////////////////////////////////////////////
+
+    private final BiFunction<Variable, Integer, Integer> inc = (v, i) -> i + 1;
+
+    final void plug(Variable var) {
+        if (!observed.containsKey(var)) {
+            observed.put(var, 1);
+            var.addMonitor(this);
+        } else {
+            observed.computeIfPresent(var, inc);
```
ArthurGodet: Why not call directly `observed.compute(var, inc)` ? `var` is a key here, and the associated value is not null "by construction" of the `observed` map

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/DomOverWDeg.java
```diff
@@ -99,15 +215,92 @@ final int remapInc() {
...
+        }
+    }
+
+    private final BiFunction<Variable, Integer, Integer> dec = (v, i) -> i - 1;
+
+    private void unplug(Variable var) {
+        assert observed.containsKey(var);
+        Integer obs = observed.computeIfPresent(var, dec);
```
ArthurGodet: Same remark, why not call directly `observed.compute(var, inc)` ?

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/variables/AbstractCriterionBasedVariableSelector.java
```diff
@@ -90,46 +58,25 @@ public Element(int count, int w0, int w1) {
...
-    };
-
     public AbstractCriterionBasedVariableSelector(V[] vars, long seed, int flush) {
         this.random = new java.util.Random(seed);
         this.solver = vars[0].getModel().getSolver();
         this.environment = vars[0].getModel().getEnvironment();
         this.last = environment.makeInt(vars.length - 1);
         this.flushThs = flush;
+//        this.weights = new Weightable.Map<>();
+        this.weights = new Weightable.Array<>();
```
ArthurGodet: Shouldn't this depend on the maximum of the id of variables ? The Array version of Weightable seems to be a source of high memory consumption if the ids of variables are not continuous from 1 (if variables have id from 500 to 2500, `weights` will rely on an array of size 2501, but indexes 0 to 499 will never be used, which seems to be "a lot" of memory for nothing). That said, if Weightable are used only in this selector, it might not be that important

cprudhom: You're suggesting to add an offset as the minimum ID among all variables to reduce the size of the array.
That's a good idea but the way AbstractCriterionBasedSelector is used allows adding weight to variables not declared initially in the scope of the instance (because for example, when a propagator fails, dom/wdeg does not ignore variable out of the declared scope).
Your remark is correct but I feel like it's going to overcomplicate a structure which goal is to be simple and efficient.
I should probably consider a way for users to switch from `Array` to `Map`.

ArthurGodet: I was not suggesting to have an offset for the `Array` implementation of `Weightable` (as it might be very difficult, or prone to inefficiency, due to the fact that variables in scope are dynamically concidered) but merely to have a condition that chooses between the `Array` and the `Map` implementation of `Weightable` depending on the ids of vars (even knowing the variables in scope of weights are dynamic). The idea being to be more memory-efficient. But, once again, if Weightable is used only in selectors, it might not be worth the trouble (as the restricted use will not lead to a lot of memory inefficiency)

cprudhom: In fact, using an Array is about 30% faster than using a Map.
This is mainly due to the access of an element in an array of doubles which is faster than retrieving the position of the key in an hashmap.
That said, that's why I suggest to Split `Model.nextId()` into `Model.nextVarId()` and `Model.nextPropId()`, to also (mainly) limit the memory footprint.

solver/src/main/java/org/chocosolver/util/objects/Weightable.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/**
+ * <br/>
+ *
+ * @author Charles Prud'homme
+ * @since 26/03/2025
+ */
+public interface Weightable<I extends Identity> {
```
jgFages: I understand the interest of allowing different implementations to identify the best one. However, now we know the array is faster, shouldn't we keep only the best one ?

solver/src/main/java/org/chocosolver/util/objects/IVal.java
```diff
@@ -25,4 +27,100 @@ public interface IVal {
...
+    }
+
+    /**
+     * <p>
+     * Project: choco-solver.
+     *
+     * @author Charles Prud'homme
+     * @since 30/10/2018.
+     */
+    class ArrayVal implements IVal {
```
jgFages: Here too, do we need various implementations (open question, I have no idea) ?

cprudhom: Yes we do, ActivityBased selects the accurate one based on some criterion

jgFages: Thanks! 
It is not really the purpose of the PR but as I see it is only used by ABS, I would like to suggest a few changes : 
- Renamming it to something more accurate, maybe IActivityTracker, IActivityStore...  I find IVal a bit too generic, values are manipulated all over the solver.
- Moving IVal to a specific abs package where is the activitybasedsearch in order to keep only generic and reusable objects in org.chocosolver.util.objects;
- Extracting implementations into separate files

solver/src/main/java/org/chocosolver/util/objects/IdentityToDouble.java
```diff
@@ -0,0 +1,105 @@
...
+        Arrays.fill(values, 0);
+    }
+
+    /**
+     * Return the top n identifiable objects wrt to their weight
+     *
+     * @param n number of objects to return
+     * @return a list of n identifiable objects
+     */
+    public List<Integer> getTop(int n) {
```
jgFages, line 106: The main reason for using this class instead of TDoubleArrayList is this method right ?

solver/src/main/java/org/chocosolver/util/objects/IdentityToDouble.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Return the weight associated with {@code i}
+     *
+     * @param i an identifiable object
+     * @return the weight associated with {@code i}
+     */
+    public double get(I i) {
+        ensureCapacity(i.getId() + 1);
+        return values[i.getId()];
```
ArthurGodet: It might be interesting to replace by this code (to avoid modifying/enlarging `values` just to get 0) ?

```
if (values.length < i.getId() + 1) {
    return 0.;
} else {
    return values[i.getId()];
}
```

solver/src/main/java/org/chocosolver/util/objects/IdentityToDouble.java
cprudhom: This method is used to estimate if the weights must be flushed or not.
The principle (hopefully) simple: if 90% of the top-20 most weighted variables does not change during `flushThs` restarts, all the weights are reset to 0.

jgFages (approved):

ArthurGodet (approved):

Issue #1160, opened 2025-05-13
Optimize checking of unsupported values in CouplesBitSetTable


jsimomaa:
Noticed this possible optimization while profiling my app.

Currently `PropBinAC3bitrm.propagate()` and `PropBinAC3bitrm.revise()` methods call `CouplesBitSetTable.checkUnsupportedValue()` in a loop with fixed `var` and `v`:

https://github.com/chocoteam/choco-solver/blob/master/solver/src/main/java/org/chocosolver/solver/constraints/extension/binary/PropBinAC3bitrm.java#L62-L90

I'll add a PR with a potential optimization that in our app increased the performance 3x

PR #1161, opened 2025-05-13
Optimize checking of unsupported values in CouplesBitSetTable

jsimomaa:
Optimize checking of unsupported values in CouplesBitSetTable #1160


cprudhom:
Hi

Thank you for PR.
Do you need help to make all tests pass?

In addition, did you test other algorithm for binary table?
I'm not 100% sure that AC3bitrm is the most efficient and/or in which cases it is the most relevant to use

jsimomaa:
> Hi
> 
> Thank you for PR. Do you need help to make all tests pass?
> 
> In addition, did you test other algorithm for binary table? I'm not 100% sure that AC3bitrm is the most efficient and/or in which cases it is the most relevant to use

Hi,

I just ran out of time but I'll try to check the tests shortly.

We initially used `CT+` but had some OutOfMemory issues thus we changed to `AC3bit+rm`

PR #1165, opened 2025-08-14
Replace Cumulative implementation by state-of-the-art implementation of TimeTabling and OverloadChecking

### LINKED ISSUE #1114: [BUG]Unexpected behavior when using Task with scalar

ArthurGodet:
Also fixes #1114


cprudhom:
I change the target branch from `master` to `develop`

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/Event.java
cprudhom, line 1: This file could be annotated a little more.

solver/src/test/java/org/chocosolver/solver/constraints/checker/Modeler.java
```diff
@@ -670,7 +670,7 @@ public Model model(int n, int[][] domains, THashMap<int[], IntVar> map, Object p
                 h[i] = vars[i + 3 * k];
             }
             IntVar capa = vars[vars.length - 1];
-            model.cumulative(tasks, h, capa, (boolean)parameters).post();
```
cprudhom, line 673: Ignoring this parameter should also involve the calling test, right?

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/EventPointSeries.java
cprudhom, line 1: This file could be annotated a little more.

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/Profile.java
cprudhom, line 1: This file could be annotated a little more.

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/PropagatorCapacity.java
cprudhom, line 1: This file could be annotated a little more.

cprudhom (commented):
Apart from the missing comments, everything is OK.

solver/src/main/java/org/chocosolver/solver/constraints/ISchedulingFactory.java
```diff
@@ -0,0 +1,251 @@
...
+            return ref().arithm(capacity, ">=", 0);
+        }
+        final Task[] keptTasks = new Task[tasksToKeep.size()];
+        final IntVar[] keptHeights = new IntVar[tasksToKeep.size()];
+        for (int i = 0; i < tasksToKeep.size(); ++i) {
+            keptTasks[i] = tasksToKeep.get(i);
+            keptHeights[i] = heightsToKeep.get(i);
+        }
+        if (keptTasks.length == 1) {
+            return ref().arithm(keptHeights[0], "<=", capacity);
```
jgFages: height <= capa OR duration = 0, no ?

jgFages:
The branch needs to be rebased from origin/develop

solver/src/test/java/org/chocosolver/solver/constraints/checker/Modeler.java
ArthurGodet: What calling test ?

solver/src/main/java/org/chocosolver/solver/constraints/ISchedulingFactory.java
ArthurGodet: Indeed, it should be this formula. Now the question is : is it better to use an OR constraint or directly the CUMULATIVE constraint on the only task ?

jgFages: I would expect the OR to be more efficient

ArthurGodet: I will manage it with PropagatorCapacity as it is doing exactly that (and is working with optional task)

jgFages: ok

ArthurGodet:
I had problems with git rebase, so, in order to keep a clean git history, we need to squash this PR whenever it is accepted/merged

CHANGES.md
```diff
@@ -70,23 +70,6 @@ See [milestone 5.0.0](https://github.com/chocoteam/choco-solver/milestone/40)
...
-
-4.10.16 - 12 Sep 2024
--------------------
-
-### Hotfix
-- Fix bug in `IntAffineView` (#1101)
-
-**Full Changelog**: https://github.com/chocoteam/choco-solver/compare/v4.10.15...v4.10.16
-
-
```
jgFages, line 89: is this change normal ?

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/NRJCumulFilter.java
```diff
@@ -1,100 +0,0 @@
...
-			if(d[i].getLB()>0){
-				slb[i] = s[i].getLB();
-				dlb[i] = d[i].getLB();
-				eub[i] = e[i].getUB();
-				hlb[i] = h[i].getLB();
-				assert eub[i]>slb[i];
-				sor_array[idx++] = i;
-			}
-		}
-		sorter.sort(sor_array,idx,comparator);
```
jgFages, line 76: Is this logic implemented somewhere else ? 
If not, I would like to keep it, within the new framework

CHANGES.md
ArthurGodet: Aren't these changes written twice (hence the deleted lines) ?

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/NRJCumulFilter.java
ArthurGodet: Yes, it is kept in the function energyNaive() of PropagatorCumulative. It is not activated by default (we tested on instances of the MiniZinc challenge with Charles), but it is kept for the cases where it helps.

ArthurGodet: However, if you talk about the while loop storing values in slb, hlb, eub arrays, then no it is not kept.
Do you see an interest in keeping it ?

jgFages: Ah great thank you very much I didn't see the energyNaive()

jgFages (approved):

ArthurGodet:
Important to note : the failing tests do not rely on the cumulative constraint

cprudhom:
Here are some results of the new Cumulative (`new`) in comparison to the current version (`base`) on instances from the last 6 MiniZinc challenges declaring `cumulative` constraints.

## With LCG off
image

The two instances that are faster with `base` than with `new` are :
- `gfd-schedule2_n200f5d50m40k6_10124` : 95s vs 100s 
- `model4_opt_test03` : 106s vs 154s

The difference comes from the usage of the OverloadChecking algorithm which never helps on these instances.

image


## With LCG on

image

The two instances where `base-lcg` is faster than `new-lcg` are:
- `mrcpsp_j30_29_3` : 22s vs 57s
- `mrcpsp_j30_54_2` : 12s vs 23s

I still need to understand the differences but as a recall, `base-lcg` time-based decomposes the cumulative constraint.

image

ArthurGodet:
I confirm that base-lcg is based on decomposition for the cumulative. I wonder if the two instances where base-lcg does better are on a short horizon ? That might explain the performance difference.

cprudhom:
Actually, in the benchmarks, there are 5 instances of mrcpsp_j30, with different results.


Instance | Opt | base-lcg | new-lcg
-- | -- | -- | --
mrcpsp_j30_25_5 | min | 27 - 3s - 995n - P | 27 - 1s - 2470n - P
mrcpsp_j30_29_3 | min | 32 - 22s - 24110n - U | 32 - 57s - 166551n - U 
mrcpsp_j30_33_5 | min | 48 - 52s - 7150n - P | 48 - 5s - 9829n - P
mrcpsp_j30_41_8 | min | 30 - 42s - 5822n - P | 30 - 4s - 7984n - P 
mrcpsp_j30_54_2 | min | 28 - 12s - 9537n - U | 28 - 23s - 75877n - P


**Edit 1** : I added the number of nodes  (XXXn) and the termination (P for proven, U for unproven). 
 Now, it seems that `base-lcg` is able to find solutions exploring fewer nodes, but since the decomposition can be big, it also wastes time propagating (see for instance, `mrcpsp_j30_25_5` or `mrcpsp_j30_33_5`).
This, also, mitigates `mrcpsp_j30_54_2` because `base-lcg` is not able to prove the bound, unlike `new-lcg`.

We can conclude that the decomposition provides better explanations (probably, thanks to additional variables introduced by the decomposition) - which is, as far as I known, not surprising.   
On the other hand, the global version of cumulative provides better propagation performances but suffer from weaker explanations.

Question is: can we live with that? 
<details><summary>My opinion</summary>Yes, we can</details>

cprudhom:
Can we consider this PR as ready to be merged ?

ArthurGodet:
I didn't see your edits. So it seems that mrcpsp_j30_29_3 is the only instance that is better treated by base_lcg, and since the best found solution is not proven to be optimal, I'd say that it is not that important (especially compared to the better results we get wih new_lcg for the other instances). I agree that the explanations of new_lcg seem weaker, but my first impression would be that it is not surprising (even if I can't really explain why).

For me, even if it can still be improved (especially by better detecting when to apply the OverloadChecking or not, to improve performance), I consider that the PR can be merged. Don't forget to **_squash and merge_** the PR as the git history of the branch is awful.

PR #1167, opened 2025-08-19
Changes made in preparation for the XCSP3 and MiniZinc competitions

cprudhom:
These changes include the following:
- Change to the default strategy for XCSP3
- Conditional transformation of expressions into constraints in extensions and detection of some classic expressions (see #embody() method)
- Implementation of a cache for the XCSP3 parser
- addition of a configurable RoundRobin strategy
- improvement and correction of DomOverWDeg and its variants
- improvement of structures for CompactTable
- creation of a class dedicated to preprocessing algorithms
- addition of flexibility in ParallelPortfolio
- other improvements and bug fixes


solver/src/main/java/org/chocosolver/solver/constraints/IDecompositionFactory.java
```diff
@@ -46,6 +48,55 @@
...
+     * Creates and <b>posts</b> a decomposition of an allDifferent_except constraint.
+     * <p>
+     * This constraint ensures that all variables from vars are pairwise different, except those which are in except.
+     * </p>
+     * This decomposition is based on the global cardinality constraint.
+     *
+     * @param vars   array of integer variables
+     * @param except array of values that can be excluded from the allDifferent constraint
+     */
+    default void allDifferentExceptDec2(IntVar[] vars, int[] except) {
```
jgFages: Maybe rename as "allDifferentExceptGCCDec" ?

solver/src/main/java/org/chocosolver/solver/constraints/reification/PropXgeCHalfReif.java
```diff
@@ -46,7 +46,7 @@ public int getPropagationConditions(int vIdx) {
         if (vIdx < 1) {
             return IntEventType.DECUPP.getMask();
         }
-        return IntEventType.INCLOW.getMask();
+        return IntEventType.INSTANTIATE.getMask();
```
jgFages, line 49: change made to trigger the setPassive ?

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/values/IntDomainSticky.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * A value selector that returns the last value assigned to a variable while propagating.
+ * If no value was assigned, falls back to 'mainSelector'.
+ * This value selector monitors the variables and stores the last value assigned to each variable.
+ * Then, when a value has to be selected, it checks if the last assigned value is still in the domain of the variable.
+ * If it is, it returns this value, otherwise it falls back to the main selector.
+ *
+ * @author Charles Prud'homme
+ */
+public final class IntDomainSticky implements IntValueSelector, Function<IntVar, OptionalInt>, IVariableMonitor<IntVar> {
```
jgFages, line 31: not used in round robin?

jgFages (approved):

solver/src/main/java/org/chocosolver/solver/constraints/reification/PropXgeCHalfReif.java
cprudhom: exactly

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/values/IntDomainSticky.java
cprudhom: Should be. I'll change that

solver/src/main/java/org/chocosolver/solver/constraints/IDecompositionFactory.java
cprudhom: Why GCC?
allDifferentExceptManyDec ?

solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/values/IntDomainSticky.java
cprudhom: Actually, it is already usable in RoundRobin

solver/src/main/java/org/chocosolver/solver/constraints/IDecompositionFactory.java
jgFages: because the decomposition is based on GCC, so I would understand it as a "GCC based decomposition"
but I admit that reading "ExceptGCC" might be confusing, so perhaps allDifferentExceptDecGCC would be better.
We can also stay with the allDifferentExceptDec2

solver/src/main/java/org/chocosolver/sat/Clause.java
```diff
@@ -139,6 +139,22 @@ public boolean learnt() {
         return learnt;
     }
 
+    /**
+     * Indicates if all literals but one are false in the clause.
+     * By implication, the remaining literal is the asserting literal.
+     * By contract, the literal at index 0 is the asserting literal and should not be considered.
```
jgFages: I find the combination of "literal at index 0 ... should not be considered" and "allFalseButOne" confusing. 
My first understanding is that were are checking that all literals but one are false, starting from position 1, whereas the code check that they are all false. If the code is correct, I would rephrase the doc : 
"By contract, the literal at index 0 is the asserting literal and always true."

cprudhom: Your first understanding is correct: we are checking that all literals but one are false, starting from position 1.
The loop starts at 1 (not 0) and if a literal is not false, the loop is broken and `false` is returned. 

Then, the contract is not that the literal at 0 is always true. 
In its current usage (for assertion only), it should be undefined, otherwise there is a problem somewhere (indeed, the current cannot explain an event that already occurs).

Note that if this method is called somewhere else in the code, the assertion can be wrong, though, since it must be true only when a literal is satisfied or falsified.

Its easier (from a conception perspective) to keep the method in the Clause class, but I agree that it can be renamed for clarity.

solver/src/main/java/org/chocosolver/sat/MiniSat.java
```diff
@@ -659,7 +680,7 @@ private void analyseConflict(Clause confl, TIntList out_learnt, int p, int pathC
                 int x = var(q);
                 if (!seen.get(x) && level(x) > rootlvl) {
                     assert p == litUndef || pos(var(p)) > pos(x) : "chronological inconsistency :(" + printLit(p) + " @ " + pos(var(p)) +
-                            ") is explained by a previous event (" + printLit(x) + " @ " + pos(x) + ") " + c;
+                            ") is explained by a older event (" + printLit(x) + " @ " + pos(x) + ") " + c;
```
ArthurGodet: typo: "a older" instead of "an older"

solver/src/main/java/org/chocosolver/solver/constraints/IDecompositionFactory.java
ArthurGodet: What about "allDifferentExceptDecBasedOnGCC" ? Or "allDifferentExceptDecByGCC" ?

solver/src/main/java/org/chocosolver/solver/constraints/IDecompositionFactory.java
```diff
@@ -187,6 +238,31 @@ default void cumulativeDec(IntVar[] starts, IntVar[] durations, IntVar[] heights
...
 
 
+    /**
+     * Creates a decomposition of the disjunctive constraint.
+     * This constraint ensures that no two tasks overlap.
+     * <p>
+     * If a task is optional, it can be performed with a duration of 0 and the other mandatory tasks can overlap it.
+     * Otherwise, the task must be performed and cannot be overlapped by any other task even if it has a duration of 0.
+     *
+     * @param tasks array of tasks
```
ArthurGodet, line 248: the `@param strict` is missing

solver/src/main/java/org/chocosolver/solver/constraints/IDecompositionFactory.java
```diff
@@ -187,6 +238,31 @@ default void cumulativeDec(IntVar[] starts, IntVar[] durations, IntVar[] heights
     }
 
 
+    /**
+     * Creates a decomposition of the disjunctive constraint.
+     * This constraint ensures that no two tasks overlap.
+     * <p>
+     * If a task is optional, it can be performed with a duration of 0 and the other mandatory tasks can overlap it.
+     * Otherwise, the task must be performed and cannot be overlapped by any other task even if it has a duration of 0.
```
ArthurGodet, line 246: This is not clear to me. I don't see, from the code, where the optionality of a task does intervene ? And shouldn't the optional tasks that are performed be considered as mandatory tasks ?

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
```diff
@@ -895,9 +939,34 @@ default Constraint allDifferentUnderCondition(IntVar[] vars, Condition condition
...
     default Constraint allDifferentExcept0(IntVar[] vars) {
+        if (ref().getSolver().isLCG()) {
+            return allDifferentExceptValues(vars, 0);
+        }
         return allDifferentUnderCondition(vars, Condition.EXCEPT_0, true);
     }
 
+    /**
+     * Creates an allDifferent constraint for variables that are not equal to 0.
+     * There can be multiple variables equal to 0.
```
ArthurGodet: Should not it be values instead of 0 ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsoluteLight.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public void propagate(int evtmask) throws ContradictionException {
+        absY.updateLowerBound(0, this, Reason.undef());
+        boolean loop;
+        do {
+            int l = Y.getLB();
+            int u = Y.getUB();
+            if (l >= 0) {
+                absY.updateLowerBound(l, this, lcg() ? Reason.r(Y.getMinLit()) : Reason.undef());
+                absY.updateUpperBound(u, this, lcg() ? Reason.r(Y.getMinLit(), Y.getMaxLit()) : Reason.undef());
```
ArthurGodet, line 90: Why is there `Y.getMinLit()` in the explanation ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsoluteLight.java
```diff
@@ -0,0 +1,104 @@
...
+            } else if (u <= 0) {
+                absY.updateLowerBound(-u, this, lcg() ? Reason.r(Y.getMaxLit()) : Reason.undef());
+                absY.updateUpperBound(-l, this, lcg() ? Reason.r(Y.getMaxLit(), Y.getMinLit()) : Reason.undef());
```
ArthurGodet, line 93: Why is there `Y.getMaxLit()` in the explanation ?

ArthurGodet (changes requested):

solver/src/main/java/org/chocosolver/solver/constraints/IDecompositionFactory.java
cprudhom: Following the MiniZinc definition (for which this decomposition has been written) a task with 0-duration is considered as optional, and can be overlapped.
But, if strict is set to true, a 0-duration task cannot be overlapped.

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsoluteLight.java
cprudhom: Because l (Y.getLB()) is >=0, which could happen due to filtering of Y lower bound.

Consider the case where Y = [-10, 5], then |Y| = [0,10]
If an event set Y to [2,5], then |Y| = [2,5] and the modification of |Y| upper bound is due to the current upper bound of Y and the fact the lower bound of Y changes (to a value >= 0)

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsoluteLight.java
cprudhom: Same reason as previous comment

solver/src/main/java/org/chocosolver/solver/constraints/IDecompositionFactory.java
ArthurGodet: Ok this is clearer for the definition and I agree on the documentation.

I still would precise that optionality is treated as the possibility of the task to have a null-duration (as it is the only criteria taken into account in the code wrt optionality of tasks)

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsoluteLight.java
ArthurGodet: I would have expected the explanation to therefore be `Reason.r(Y.getLELit(-1), Y.getMaxLit())` but it would only work for enumerated-domains' variables. Maybe we could rely on the `getNegGeqLit` and `getNegLeqLit` functions I used in PropagatorCumulative (by making them static functions in Reason ?)

PR #1107, opened 2024-10-22
Set Expressions (with Chainable Format) for Arithmetic and Relational Operations

gadavidd:
**Set Expressions (with Chainable Format) for Arithmetic and Relational Operations**

I've been working with `SetVar` in `choco-Solver`. Unlike other types of variables, such as `RealVar` or `IntVar`, there was no direct way to compose operations such as equals (eq), not equals (ne), contains, not contains (nc), subset (subSet), union, or intersection directly in the model definition. While propagators allow for constructing such constraints, it was not as intuitive as it is for other variable types like `IntVar` and `RealVar`. These features are extremely useful in facilitating the composition of constraints.

Thus, I've worked on adding this feature for `SetVar`, based on the operations already available for other variable types. For example:

```java
realVar.eq(3).post();
```

However, such expressions were not possible for `SetVar`:

```java
setVar.eq(1, 2, 3).post();  // Not available... until now!
```

**Improvement**

I've implemented a new expression framework for `SetVar`, allowing the use of familiar operations directly on set variables. This change introduces support for the following operations:

Relational Operations:

- `eq` (Equals): `setA.eq(1, 2, 3)` or `setA.eq(setB)`.
- `ne` (Not Equals): `setA.ne(1, 2)` or `setA.ne(setB)`.
- `contains`: `setA.contains(1, 2)`.
- `nc` (notContains): `setA.notContains(1, 2)`.
- `subSet`: `setA.subSet(1, 2, 3)` or `setA.subSet(setB)`.

Arithmetic Set Operations:

`union`: `setA.union(setB)`
`intersection`: `setA.intersection(setB)`

This enhancement allows for the expression of complex set relations with simplified syntax, enabling combinations such as:

```java
setA.eq(setB.union(setC.intersection(setD))).post();
```

**Example Usage:**
```java
Model model = new Model();

// Define set variables
SetVarImplArExpression setA = new SetVarImplArExpression(model, "setA", new int[]{}, new int[]{0, 1, 2, 3});
SetVarImplArExpression setB = new SetVarImplArExpression(model, "setB", new int[]{}, new int[]{1, 2, 3});
SetVarImplArExpression setC = new SetVarImplArExpression(model, "setC", new int[]{}, new int[]{2, 3, 4, 5});

// Compose set expressions
setA.eq(setB.union(setC)).post();
model.getSolver().propagate();

// Assert the expected values in setA
[...]

PR #1170, opened 2025-08-28
Set Expressions (with Chainable Format) for Arithmetic and Relational Operations 

cprudhom:
@gadavidd I wasn't able to push to your branch so I created a new one, namely `gadavidd-set-expression-improvements`.
It is based on your work (#1107), rebased upon the current version of `develop`.
It also includes some modifications to align the structure with what was done for int expressions.

Other solution: add my last commit to your branch and update your PR (in that case, authorship is clear).
Let me know your choice.


gadavidd:
@cprudhom I believe that the first solution is more suitable for choco-solver. As far as I'm concerned, this approach is fine. In any case, I agree with whatever you think is best. The important thing is that we continue to contribute to the development of choco-solver.

cprudhom:
Ok, so in that case, I'll cherry-pick the files and create a new PR.

PR #1171, opened 2025-08-29
Set Expressions (with Chainable Format) for Arithmetic and Relational Operations

cprudhom:
This PR takes the work from @gadavidd (#1107)  and extracts the 4 impacting commits and places them back on the up-to-date `develop` branch.


solver/src/main/java/org/chocosolver/solver/expression/discrete/set/BiReSetExpression.java
```diff
@@ -0,0 +1,89 @@
...
+        SetVar xSet = x.setVar();
+        SetVar ySet = y.setVar();
+        switch (op) {
+            case SUBSET:
+                return model.subsetEq(xSet, ySet);
+            case EQ:
+                return model.allEqual(xSet, ySet);
+            case NE:
+                return model.allDifferent(xSet, ySet);
+            case CONTAINS:
```
jgFages, line 79: Why having both SUBSET and CONTAINS ?
I think I prefer subset

solver/src/main/java/org/chocosolver/solver/expression/discrete/set/BiReSetExpression.java
```diff
@@ -0,0 +1,89 @@
...
+                return model.subsetEq(ySet, xSet);
+            case NOT_CONTAINS:
+                return model.disjoint(xSet, ySet);
```
jgFages, line 82: if contains(x, y) means that y is a subset of x
I would expect not_contains(x, y) to mean that y is not a subset of x, which does not mean that both sets are disjoint. They might have one value in common.

solver/src/main/java/org/chocosolver/solver/expression/discrete/set/ArSetExpression.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Checks if this set is different from a given set of integers.
+     *
+     * @param y the integer values to compare.
+     * @return a boolean expression representing inequality.
+     */
+    default ReExpression ne(int... y) {
+        return new BiReSetExpression(SetOperator.NE, this, this.getModel().setVar(Arrays.stream(y).toArray()));
+    }
```
ArthurGodet, line 162: I would have relied on the methods creating the `BiReSetExpression` from a `ArSetExpression`, once the `SetVar` created from `y` values, instead of relying on `new BiReSetExpression()` in these methods.

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
```diff
@@ -0,0 +1,380 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+
+public class SetVarExpressionTest {
+
+    /**
+     * Basic
+     * ============= Relacional ============= values and sets
```
ArthurGodet, line 24: typo: "Relational"

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
```diff
@@ -0,0 +1,380 @@
...
+     * <p>
+     * Simple and Compound -> a.eq((b.intersection(c.union(d)).union(e.intersection(f))).intersection(g)
+     * <p>
+     * Another functions
+     * setCard - set Cardinality
+     * notEmpty - a set can't empty
+     * Empty
+     */
+
+    @Test
```
ArthurGodet, line 45: Should not it be a test from TestNG (more especially one of the "1s" category) ?

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
```diff
@@ -0,0 +1,380 @@
...
+
+        model.getSolver().propagate();
+
+        model.notMember(0, setA).post();
+        model.notMember(1, setA).post();
+        model.notMember(3, setA).post();
+
+        model.getSolver().propagate();
+
+        assertEquals(2, setA.getValue().toArray()[0]);
```
ArthurGodet, line 61: I would finish this test by testing that we have a ContradictionException when trying to remove value 2, as the set should not be empty

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
```diff
@@ -0,0 +1,380 @@
...
+    public void simpleSubSetTest01() throws ContradictionException {
+        Model model = new Model();
+        SetVar setA = model.setVar("setA", new int[]{}, new int[]{0, 1, 2, 3, 4, 5, 6, 7, 8, 9});
+        setA.subSet(1, 2, 3, 4, 5).post();
+
+        model.getSolver().propagate();
+
+        assertArrayEquals(new int[]{1, 2, 3, 4, 5}, setA.getUB().toArray());
+
+        setA.notEmpty().post();
```
ArthurGodet, line 166: Is this relevant for the test ?

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
```diff
@@ -0,0 +1,380 @@
...
+        SetVar setA = model.setVar("setA", new int[]{}, new int[]{0, 1, 2, 3, 4, 5, 6, 7, 8, 9});
+        SetVar setB = model.setVar("setB", new int[]{}, new int[]{1, 2, 3, 4, 5});
+        setA.subSet(setB).post();
+
+        model.getSolver().propagate();
+
+        assertArrayEquals(new int[]{1, 2, 3, 4, 5}, setA.getUB().toArray());
+        assertArrayEquals(new int[]{1, 2, 3, 4, 5}, setB.getUB().toArray());
+
+        setA.notEmpty().post();
```
ArthurGodet, line 278: Is this relevant for the test ?

ArthurGodet (changes requested):

solver/src/main/java/org/chocosolver/solver/expression/discrete/set/BiReSetExpression.java
ArthurGodet: Not necessarily. How would you express the following constraint otherwise ? "x should not contain the values of y ?"

jgFages: I would use the term disjoint for that. It is much clearer.
"Should not contain the values of y" is confusing because a value of a set variable is a set. 
Let say X has 2 solutions : {2,3} and {2,5,7}. If Y = {2}, then it does not contain the values of X. Yet, I guess it is not what you meant, you want both variables to have disjoint values.
Also, to contain is a directed concept (A contains B is different from B contains A) whereas disjointness is not.

ArthurGodet: I see what you mean. Therefore, I agree that DISJOINT should be a better name than NOT_CONTAINS

gadavidd: I agree as well. When I wrote this term, I was thinking from a user’s perspective. In that context, “contains” and “not contains” are more commonly used, but they introduce ambiguity. Using “subset” and “disjoint” makes the intended behavior much clearer.

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
gadavidd: `NotEmpty` is used to prevent `setA` from being empty, but I don’t think this is very relevant here. The test would produce the same result without `NotEmpty`.

cprudhom:
@gadavidd Even if I created the PR and revamp the code a little, this is your work. 
Would you mind to consider the comments here and integrate them in a new commit on this branch?

solver/src/main/java/org/chocosolver/solver/expression/discrete/set/BiReSetExpression.java
gadavidd: CONTAINS was originally added as a semantic counterpart to NOT_CONTAINS. However, it does not add much value internally in ArSetExpression, since the same relation can be expressed with SUBSET by swapping the operands. I changed CONTAINS and NOT_CONTAINS to SUBSET and DISJOINT, respectively.

gadavidd:
@cprudhom Sure, sorry for the delay; I had some personal issues, but I am now reviewing the comments and will integrate the required changes in a new commit on this branch.

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
gadavidd: Portuguese, sorry

solver/src/main/java/org/chocosolver/solver/expression/discrete/set/ArSetExpression.java
gadavidd: Thanks, I agree. I updated the int... overloads to create the `SetVar` first and delegate to the corresponding ArSetExpression-based methods.

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
gadavidd: You're right, this is not relevant. I removed it.

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
gadavidd: it done

examples/src/test/java/org/chocosolver/examples/set/expression/SetVarExpressionTest.java
gadavidd: Yes, I changed it

gadavidd:
@cprudhom @ArthurGodet @jgFages I addressed the review comments, but I do not have permission to push to dev_setexpr. I pushed the changes to my fork and opened PR #1196 with the updates.

https://github.com/chocoteam/choco-solver/pull/1197

cprudhom:
I think the PR is ready to be merged.
@jgFages @ArthurGodet any additional comments?

ArthurGodet (approved):

jgFages (approved):

PR #1173, opened 2025-10-16
Explained Bimodal AllDifferent

NelusBreizh:
This code is related to the algorithm described in the following paper presented at IJCAI 2025:

https://www.ijcai.org/proceedings/2025/291


solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/algo/AlgoAllDiffBimodal.java
```diff
@@ -0,0 +1,744 @@
...
+            if (!matching.inMatchingU(var)) {
+                valuesDynamic.refill();   // We refill the list with the recently removed elements, instead of recreating it from scratch
+                int val = augmentingPath(var);
+                if (val != fail) {
+                    augmentMatching(val);
+                }
+                else {
+                    // It is not possible to get a maximum matching --> the constraint can not be satisfied
+                    //NOTE: BUG FOUND ---> when an exception is raised, the refill operation is not call and the structure of the tracking list is broken
+                    //NOTE: solution ---> integrate the refill operation in the generateFailureExplanation() method, before raising the exception
```
jgFages: Is the bug still there ? 
If yes, it should be corrected.
If not, the comments should be removed

solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/algo/AlgoAllDiffBimodal.java
```diff
@@ -0,0 +1,744 @@
...
+        this.pruned = false;
+        updateDynamicStructuresOpening(false);
+        boolean foundMatching = findMaximumMatching();
+        if(foundMatching) {
+            filter();
+            updateDynamicStructuresEnding();
+        }
+        //NOTE: don't need anymore because of generateFailureExplanation
+//        else {
+//            vars[0].instantiateTo(vars[0].getLB() - 1, aCause);
```
jgFages: remove unused code & comments

solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/AllDifferent.java
```diff
@@ -37,10 +45,16 @@ public AllDifferent(IntVar[] vars, String type) {
...
+        if(!Objects.equals(OPTION_CONSISTENCY, "DEFAULT")){
+            consistency = OPTION_CONSISTENCY;
+        }
+        boolean propInst = !Objects.equals(OPTION_PROPINST, "false") && !Objects.equals(OPTION_PROPINST, "FALSE");
+
         Model model = VARS[0].getModel();
         if (model.getSolver().isLCG()) {
             String message = "";
             if (consistency.equals("AC") || consistency.equals("AC_ZHANG")) {
+                //TODO: put AC_TUNED as default ?
```
jgFages: any thoughts @cprudhom ?

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
```diff
@@ -846,7 +846,7 @@ default Constraint times(IntVar X, IntVar Y, IntVar Z) {
      * @param vars list of variables
      */
     default Constraint allDifferent(IntVar... vars) {
-        return allDifferent(vars, "DEFAULT");
+        return allDifferent(vars, "AC_COMPLEMENT");
```
jgFages: I would keep "DEFAULT" as the default option, unless the idea is to remove it?

solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/AllDifferent.java
```diff
@@ -17,16 +17,24 @@
...
+import java.util.Objects;
 
 /**
  * Ensures that all variables from VARS take a different value.
  * The consistency level should be chosen among "AC", "BC", "FC" and "DEFAULT".
  */
 public class AllDifferent extends Constraint {
 
+    public static String OPTION_CONSISTENCY = "DEFAULT";
+    public static String OPTION_PROPINST = "DEFAULT";
```
jgFages: why having static non-final values here ?

cprudhom:
# REGIN vs COMPLEMENT 
## on COP 
### without LCG
radar_cop_nox
cactus_cop_nox

### with LCG
radar_cop_x
cactus_cop_x

## on CSP
### without LCG
radar_csp_nox
cactus_csp_nox


### with LCG

radar_csp_x
cacstu_csp_x


# COMPLEMENT vs CLASSIC vs TUNED

## on COP
### without LCG
radar_cop_ad
cactus_cop_ad

### with LCG
radar_cop_adx
cactus_cop_adx


## on CSP
### without LCG
radar_csp_ad
cactus_csp_ad

### with LCG
radar_csp_adx
cactus_csp_adx

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
cprudhom: My opinion is that the DEFAULT algorithm should target to AC_TUNED or AC_COMPLEMENT.
This also answers to the previous question/

jgFages: Regarding results (thanks), it seems to me that Regin's algorithm remain better for COP with LCG, isn't it?
Do we plan to set LCG on by default ? If yes, Regin may be a better default option for AllDifferent. 
If not, I agree.

Nevertheless, all this assumes that AC is the best option. 
What about BC and the mix we have/had BC+proba_AC ?

cprudhom: I do not agree : Regin's algo is not better for COP + LCG. 
In 15 out of 20 cases COMPLEMENTx is better and in 11 out of 18 it is faster.

I can extract the statistics for the related problems, that would be easier to conclude

cprudhom: That said, that's true that the comparison with BC and DEFAULT is missing.

I will launch something during the week-end

jgFages: Thank you, I misread the chart

solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/AllDifferent.java
```diff
@@ -37,10 +45,16 @@ public AllDifferent(IntVar[] vars, String type) {
     }
 
     private static Propagator[] createPropagators(IntVar[] VARS, String consistency) {
+        if(!Objects.equals(OPTION_CONSISTENCY, "DEFAULT")){
+            consistency = OPTION_CONSISTENCY;
+        }
+        boolean propInst = !Objects.equals(OPTION_PROPINST, "false") && !Objects.equals(OPTION_PROPINST, "FALSE");
```
ArthurGodet: You can use the equalsIgnoreCase() of the String `OPTION_PROPINST` here.

solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/AllDifferent.java
```diff
@@ -37,10 +45,16 @@ public AllDifferent(IntVar[] vars, String type) {
     }
 
     private static Propagator[] createPropagators(IntVar[] VARS, String consistency) {
+        if(!Objects.equals(OPTION_CONSISTENCY, "DEFAULT")){
+            consistency = OPTION_CONSISTENCY;
+        }
```
ArthurGodet: It is very awkward to modify the method's parameter with a value of what appers to be a static variable, depending on such static variable's value.
As a matter of fact, `consistency` should not be changed inside this method, as it is a variable used to parameterise the propagator's filtering algorithm.

Maybe we should consider using an Enum for parameterising the Alldifferent's filtering algorithm.

solver/src/main/java/org/chocosolver/util/objects/BipartiteMatching.java
```diff
@@ -0,0 +1,154 @@
...
+    //  * Returns the array of the matched vertices of V
+    //  * Warning: If all vertices of U are not matched it will contain some UNMATCHED values, use it wisely
+    //  */
+    // public int[] getMatchedU() {return matchingV;}
+
+    // /**
+    //  * Returns the array of the matched vertices of V
+    //  * Warning: If all vertices of V are not matched it will contain some UNMATCHED values, use it wisely
+    //  */
+    // public int[] getMatchedV() {return matchingU;}
```
ArthurGodet: I think this unused code should be removed

solver/src/main/java/org/chocosolver/util/objects/BipartiteMatching.java
```diff
@@ -0,0 +1,154 @@
...
+public class BipartiteMatching {
+    private final int minU;
+    private final int maxU;
+    private final int sizeU; 
+    private final int minV;
+    private final int maxV;
+    private final int sizeV;
+    private final int UNMATCHED;
+    private int[] matchingU;
+    private int[] matchingV;
```
ArthurGodet: Why are these arrays not final ?

solver/src/main/java/org/chocosolver/util/objects/BipartiteMatching.java
```diff
@@ -0,0 +1,154 @@
...
+ */
+
+public class BipartiteMatching {
+    private final int minU;
+    private final int maxU;
+    private final int sizeU; 
+    private final int minV;
+    private final int maxV;
+    private final int sizeV;
+    private final int UNMATCHED;
```
ArthurGodet: Should be refactor to `unmatched` if it is not a constant

solver/src/main/java/org/chocosolver/util/objects/BipartiteMatching.java
```diff
@@ -0,0 +1,154 @@
...
+        String printedString = "";
+        for (int u = minU; u <= maxU; u++) {
+            if (inMatchingU(u)) {System.out.println("x" + u + " <--> " + getMatchU(u));}
+        }
+        return printedString;
+    }
+
+    public boolean equals(BipartiteMatching otherMatching) {
+        if (sizeU != otherMatching.getsizeU() || sizeV != otherMatching.getsizeV()) {return false;}
+        for (int u = minU; u <= sizeU; u++) {
```
ArthurGodet: Shouldn't it be `maxU` instead of `sizeU` ?

solver/src/main/java/org/chocosolver/util/objects/BipartiteMatching.java
ArthurGodet, line 1: Is using this data structure more performant than using two TIntIntHashMap ?

solver/src/main/java/org/chocosolver/util/objects/TrackingList.java
```diff
@@ -0,0 +1,209 @@
...
+ * @author Sulian Le Bozec-Chiffoleau
+ * @since 17 Oct. 2024
+ */
+
+public class TrackingList {
+    private int[] successor;
+    private int[] predecessor;
+    private int[] stackRemoved;
+    private int topRemoved;
+    private boolean[] present;
```
ArthurGodet: The arrays should be final too

solver/src/main/java/org/chocosolver/util/objects/TrackingList.java
```diff
@@ -0,0 +1,209 @@
...
+    public void removeFromUniverse(int e, IEnvironment env) {
+        assert(topRemoved == 0);
+        assert(isPresent(e));
+        int i = convertToIndex(e);
+        int pi = predecessor[i];
+        int si = successor[i + 1];
+        successor[pi + 1] = si;
+        predecessor[si] = pi;
+        universeSize--;
+        present[i] = false;
```
ArthurGodet: Should call `removeFromUniverse(e)` to avoid copy-pasting code

solver/src/main/java/org/chocosolver/util/objects/TrackingList.java
```diff
@@ -0,0 +1,209 @@
...
+package org.chocosolver.util.objects;
+
+import org.chocosolver.memory.IEnvironment;
+
+/**
+ * The Tracking list is similar to a doubly linked list
+ * for which each element has a predecessor and a successor, implemented by arrays
+ * An artifical source node is added at the beginning of the list
+ * An artificial sink node is added at the end of the list
+ * The tracking list allows to call the functions TrackPrev and TrackNext that are specific to this data structure
```
ArthurGodet: TrackPrev and TrackNext are not in the class. The Javadoc should be updated

solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/algo/AlgoAllDiffBimodal.java
```diff
@@ -0,0 +1,744 @@
...
+            case PARTIAL:
+//                return vars[var].getDomainSize() < complementSCC.getSize();
+                return Math.max((vars[var].getUB() - vars[var].getLB() + 63) >> 6, vars[var].getDomainSize()) < complementSCC.getSize(); // More suited to BitSet domain representation
+            case TUNED:
+//                return vars[var].getDomainSize() < complementSCC.getSize();
+                return Math.max((vars[var].getUB() - vars[var].getLB() + 63) >> 6, vars[var].getDomainSize()) < complementSCC.getSize(); // More suited to BitSet domain representation
+            default:
+                return true;
+        }
+    }
```
ArthurGodet: These functions should be refactored as most of their content is identical

ArthurGodet (changes requested):

solver/src/main/java/org/chocosolver/util/objects/BipartiteMatching.java
NelusBreizh: The BipartiteMatching class represents a bipartite matching between two sets U and V that are intervals, and thus defined by their bounds.

Thus, using arrays as done in the class is both memory and time efficient by design.

It may be interesting to compare the performances with TIntIntHashMap.

cprudhom:
# New results

## DEFAULT vs AC_TUNED
### COP
#### without LCG
image
image

The instance where DEFAULT is better than AC_TUNED is:
Instance | Opt | DEFAULT | AC_TUNED | BC
-- | -- | -- | -- | --
model_data2_6_15 | max | 7523 - 33s | 7561 - 575s | 7523 - 33s 

But, this is due to BC (last column).


#### With LCG
image
image

The instances where DEFAULT is better than AC_TUNEDx are: 

Instance | Opt | DEFAULTx | AC_TUNEDx | BCx
-- | -- | -- | -- | -- 
ctw_A041 | min | 2531504 - 249s | 2531506 - 415s | 2406504 - 305s
mapf_ins_g16_p10_a20 | min | 318 - 595s | 321 - 597s | 316 - 598s
mapf_ins_g16_p10_a30 | min | 515 - 572s | 518 - 597s | 515 - 583s
mapf_ins_g24_p20_a10 | min | 202 - 578s | 203 - 570s | 202 - 577s
model_data2_6_15 | max | 11654 - 345s | 9654 - 409s | 272 - 167s
sudoku_opt_sudoku_p23 | min | -227 - 596s | (0) | -227 - 594s

Here again, the good results are due to BC.

So, I suppose that AC_REGIN can be replaced by AC_TUNED in DEFAULT.


### CSP
#### Without LCG
image
image

#### With LCG
image
image

jgFages:
Great thanks!

cprudhom:
# AC_TUNED vs AC_TUNED+adaptive

AC_TUNED (`base`) is compared to DEFAULT (`adapt`), aka adaptive alldifferent, where the AC algorithm is set to AC_TUNED.

## without lcg
### COP
image
image

### CSP
image
image


## with lcg
### COP
image
image

### CSP
image
image


I won't go into too much details, but, when `adapt` is slower than `base` the rate of increase is equal to 4.38 on average (with a std  = 5.62). On the other hand, when `base` is slower: mean = 7.24 and std = 21.35.

So, in conclusion, I think we can replace AC_ZHAND by AC_TUNED in DEFAULT.

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
cprudhom: I switch back to "DEFAULT" since the logic should be managed by the AllDifferent class

solver/src/main/java/org/chocosolver/solver/constraints/nary/alldifferent/AllDifferent.java
cprudhom: This was done for testing purpose. Now that the code is stable, we can remove it

ArthurGodet (approved):

cprudhom:
There is a bug in CircuitTest.test4():

```bash
java.lang.UnsupportedOperationException: unexpected situation

	at org.chocosolver.solver.constraints.nary.circuit.PropNoSubtour.varInstantiated(PropNoSubtour.java:126)
	at org.chocosolver.solver.constraints.nary.circuit.PropNoSubtour.propagate(PropNoSubtour.java:100)
```

It occurs when the alldiff constraint embedded in circuit is set to AC_TUNED, AC_COMPLEMENT, AC_PARTIAL or AC_CLASSIC. 
I still need to check what is going on.

cprudhom:
This is related to the fact that the variables are bounded.
At a given node, the domains are:
```
0 = {IntervalIntVarImpl@2496} "x[0] = 2"
1 = {IntervalIntVarImpl@2497} "x[1] = [1,5]"
2 = {IntervalIntVarImpl@2498} "x[2] = [1,4]"
3 = {IntervalIntVarImpl@2499} "x[3] = [1,5]"
4 = {IntervalIntVarImpl@2500} "x[4] = [1,5]"
5 = {IntervalIntVarImpl@2501} "x[5] = 0"
```
and the decision `x[2]=1` is applied, which leads to:
```
0 = {IntervalIntVarImpl@2496} "x[0] = 2"
1 = {IntervalIntVarImpl@2497} "x[1] = [2,4]"
2 = {IntervalIntVarImpl@2498} "x[2] = 1"
3 = {IntervalIntVarImpl@2499} "x[3] = [2,5]"
4 = {IntervalIntVarImpl@2500} "x[4] = [2,5]"
5 = {IntervalIntVarImpl@2501} "x[5] = 0"
```

Which is not correct, it should be:
```
0 = {IntervalIntVarImpl@2489} "x[0] = 2"
1 = {IntervalIntVarImpl@2490} "x[1] = [3,4]"
2 = {IntervalIntVarImpl@2491} "x[2] = 1"
3 = {IntervalIntVarImpl@2492} "x[3] = [3,5]"
4 = {IntervalIntVarImpl@2493} "x[4] = [3,5]"
5 = {IntervalIntVarImpl@2494} "x[5] = 0"
```

@NelusBreizh can you have a look at this specific case ?
It seems that something is wrong here.

SulianLBC:
The propagator did not take into account the possiblity that some variables are defined by their bounds, and thus not enumerated.

The problem arises because of our dynamic management of the list of values, which implictly represents the graph. In the case where all variables are enumerated, we can safely remove from the list of values the ones that are associated with variables instantiated either during or before the call to the propagator, as those values do not belong to the domain of any variable that remains in the graph.

However, when some variables are bounded, the corresponding pruning may not have been done, if the value to remove is inside the interval.

In that case, the value was still removed from the list of values despite being present in the domain of the bounded variable. Therefore, in future calls to the propagator, the concernend value was not seen by the algorithm and could not be removed from the bounded domain although it would have been possible (if it corresponds to one of the bounds).

To fix this issue, it is sufficient to assess that a value is present in the domain of no bounded variable before removing it from the dynamic list of values.

cprudhom:
The last error in the test suite is related to the dynamic search strateg.
The new allDifferent algorithm does not filter values in the same order as the previous version.
So, it may lead to different conflicts and thus explores the search space in the same way.
When I set a static search strategy, there is no error anymore. 
I'll fix the test value and commit.

PR #1178, opened 2026-01-08
Performance improvement in PropXPlusYEqZ

jgFages:


solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsolute.java
```diff
@@ -96,13 +96,13 @@ public void propagate(int varIdx, int mask) throws ContradictionException {
...
-            } else if (Y.hasEnumeratedDomain()) {
+            } else if (Y.hasUnfixedEnumeratedDomain()) {
                 int val = X.getValue();
                 Y.updateLowerBound(-val, this, lcg() ? Reason.r(X.getValLit()) : Reason.undef());
                 Y.updateUpperBound(val, this, lcg() ? Reason.r(X.getValLit()) : Reason.undef());
                 val--;
-                for (int v = -val; v <= val; v = Y.nextValue(v)) {
-                    Y.removeValue(v, this, lcg() ? Reason.r(X.getValLit()) : Reason.undef());
+                if (val >= 0) {
+                    Y.removeInterval(-val, val, lcg() ? Reason.r(X.getValLit()) : Reason.undef());
```
cprudhom: removeInterval is not compatible with lcg.

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsolute.java
```diff
@@ -124,10 +124,9 @@ private void setBounds() throws ContradictionException {
         int min = X.getLB();
         Y.updateLowerBound(-max, this, lcg() ? Reason.r(X.getMaxLit()) : Reason.undef());
         Y.updateUpperBound(max, this, lcg() ? Reason.r(X.getMaxLit()) : Reason.undef());
-        for (int v = 1 - min; v <= min - 1; v = Y.nextValue(v)) {
-            Y.removeValue(v, this, lcg() ? Reason.r(X.getMinLit()) : Reason.undef());
+        if (1 - min <= min -1) {
+            Y.removeInterval(1 - min, min - 1, lcg() ? Reason.r(X.getMinLit()) : Reason.undef());
```
cprudhom: same here.
We should either accept not to use removeInterval or use it only when LCg is disabled

cprudhom:
Do you observe any performance improvement ?

jgFages:
For PropAbsolute, using removeInterval instead of removeValue enables to double the nb nodes per second

parsers/src/test/java/org/chocosolver/flatzinc/PerformanceTest.java
cprudhom, line 1: I think calling `System.gc()` is useless.
Did you see any differences ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsolute.java
```diff
@@ -96,13 +96,13 @@ public void propagate(int varIdx, int mask) throws ContradictionException {
             if (varIdx == 1) {
                 X.instantiateTo(Math.abs(Y.getValue()), this, lcg() ? Reason.r(Y.getValLit()) : Reason.undef());
                 setPassive();
-            } else if (Y.hasEnumeratedDomain()) {
+            } else if (Y.hasUnfixedEnumeratedDomain()) {
```
cprudhom: The condition seems correct but I think a little comment would help here.

parsers/src/test/java/org/chocosolver/flatzinc/PerformanceTest.java
jgFages: I was just hoping it would be sufficient to solve the out of memory error in the github instance (it wasn't).

solver/src/main/java/org/chocosolver/solver/constraints/binary/PropAbsolute.java
jgFages: I will move back to the original condition actually

cprudhom (approved):

PR #1179, opened 2026-01-22
Jg/iterator set

jgFages:


solver/src/main/java/org/chocosolver/util/objects/setDataStructures/linkedlist/Set_LinkedList.java
```diff
@@ -137,67 +152,54 @@ public int min() {
...
+	public int[] toArray() {
+		TIntArrayList copy = new TIntArrayList();
+		IntCell current = first;
+		while (current != null) {
+			copy.add(current.element);
+			current = current.next;
+		}
+		return copy.toArray();
+	}
+
```
ArthurGodet, line 182: Why not use the size attribute to directly create an array with the right size which will be filled within the while loop ?

ArthurGodet (changes requested):

ArthurGodet (approved):

solver/src/test/java/org/chocosolver/solver/constraints/graph/connectivity/ReifiedConnectivityTest.java
```diff
@@ -0,0 +1,807 @@
...
+        int free = 0;
+        int itf = 0;
+
+        while (solver.solve()) {
+            if(isFree.getValue() == 1)
+                free++;
+            else
+                itf++;
+        }
+        System.out.println("free: " + free +" itf: "+ itf);
```
cprudhom: I think we should add at least one test here or remove the class

cprudhom (changes requested):
At least one test in ReifiedConnectivityTest should be added

PR #1181, opened 2026-02-06
Revamp Settings

cprudhom:
This PR makes the Settings class immutable and declares a SettingsBuilder class that allows you to specify parameters in different ways and ultimately build a Settings instance.
For simplicity's sake, the RegParser class extends SettingsBuilder, avoiding the duplication of certain parameters. 
One additional thing to do: harmonize/simplify the possible options (the names are really long and unclear).

In practice, it is now possible to execute the following code:

```java
Properties prop = new Properties();
prop.setProperty("nbMaxLearntClauses", "20");
prop.setProperty("unknown", "42");
//...
SettingsBuilder sb = SettingsBuilder.init();
sb.setNbMaxLearntClauses(10);
sb.fromProperties(prop);
sb.fromArgs(new String[]{"--nbMaxLearntClauses", "30", "-P", "not_known=-42"});
Assert.assertEquals(sb.getNbMaxLearntClauses(), 30);
  Model m = new Model(sb.build());
  // ...
```
or directly in CLI with ChocoFZN and ChocoXCSP:
```bash
java -Dunknown=42 -cp choco.jar --nbMaxLearntClauses 30 ...
```

Furthermore, this should simplify the hyperparameterization of the solver...

### DIFF
--- a/parsers/src/main/java/org/chocosolver/parser/flatzinc/ast/FConstraint.java
+++ b/parsers/src/main/java/org/chocosolver/parser/flatzinc/ast/FConstraint.java
@@ -408,7 +408,7 @@ public enum FConstraint {
             IntVar c = exps.get(2).intVarValue(model);
             BoolVar r = exps.get(3).boolVarValue(model);
 
-            if ((boolean) model.getSettings().get("adhocReification").orElse(false)) {
+            if (Boolean.getBoolean(model.getSettings().get("adhocReification").orElse("false"))) {
                 if (bs.length == 1) {
                     if (bs[0].isInstantiated() || c.isInstantiated()) {
                         IntVar x;
@@ -665,7 +665,7 @@ public enum FConstraint {
             if ((a.getTypeAndKind() & Variable.KIND) == Variable.BOOL && ((b.getTypeAndKind() & Variable.KIND) == Variable.BOOL)) {
                 model.addClausesBoolIsNeqVar((BoolVar) a, (BoolVar) b, r);
             } else {
-                if ((boolean) model.getSettings().get("adhocReification").orElse(false)) {
+                if (Boolean.getBoolean(model.getSettings().get("adhocReification").orElse("false"))) {
                     if (a.isInstantiated() || b.isInstantiated()) {
                         IntVar x;
                         int c;
--- a/parsers/src/main/java/org/chocosolver/parser/flatzinc/ast/FConstraint.java
+++ b/parsers/src/main/java/org/chocosolver/parser/flatzinc/ast/FConstraint.java
@@ -408,7 +408,7 @@ public enum FConstraint {
             IntVar c = exps.get(2).intVarValue(model);
             BoolVar r = exps.get(3).boolVarValue(model);
 
-            if (Boolean.getBoolean(model.getSettings().get("adhocReification").orElse("false"))) {
+            if (Boolean.getBoolean(model.getSettings().get(ADHOC_REIFICATION).orElse("false"))) {
                 if (bs.length == 1) {
                     if (bs[0].isInstantiated() || c.isInstantiated()) {
                         IntVar x;
@@ -665,7 +665,7 @@ public enum FConstraint {
             if ((a.getTypeAndKind() & Variable.KIND) == Variable.BOOL && ((b.getTypeAndKind() & Variable.KIND) == Variable.BOOL)) {
                 model.addClausesBoolIsNeqVar((BoolVar) a, (BoolVar) b, r);
             } else {
-                if (Boolean.getBoolean(model.getSettings().get("adhocReification").orElse("false"))) {
+                if (Boolean.getBoolean(model.getSettings().get(ADHOC_REIFICATION).orElse(Boolean.FALSE.toString()))) {
                     if (a.isInstantiated() || b.isInstantiated()) {
                         IntVar x;
                         int c;
@@ -2617,6 +2617,8 @@ public enum FConstraint {
         }
     };
 
+
+    public static final String ADHOC_REIFICATION = "adhocReification";
 
     public abstract void build(Model model, Datas datas, String id, List<Expression> exps, List<EAnnotation> annotations);
 
--- a/solver/src/main/java/org/chocosolver/solver/Settings.java
+++ b/solver/src/main/java/org/chocosolver/solver/Settings.java
@@ -17,7 +17,6 @@ import org.chocosolver.solver.variables.impl.IntVarLazyLit;
 import java.util.HashMap;
 import java.util.Optional;
 import java.util.function.Consumer;
-import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
@@ -51,8 +50,6 @@ public class Settings {
 
     private final boolean sortPropagatorActivationWRTPriority;
 
-    private final int maxPropagatorPriority;
-
     private final Consumer<Model> defaultSearch;
 
     private final boolean warnUser;
@@ -83,9 +80,7 @@ public class Settings {
 
     private final boolean ibexRestoreRounding;
 
-    private final Function<Model, Solver> initSolver;
-
-    private final HashMap<String, Object> additionalSettings;
+    private final HashMap<String, String> additionalSettings;
 
     private final boolean lcg;
 
@@ -101,7 +96,6 @@ public class Settings {
         this.maxTupleSizeForSubstitution = builder.getMaxTupleSizeForSubstitution();
         this.maxSizeInMBToUseCompactTable = builder.getMaxSizeInMBToUseCompactTable();
         this.sortPropagatorActivationWRTPriority = builder.sortPropagatorActivationWRTPriority();
-        this.maxPropagatorPriority = builder.getMaxPropagatorPriority();
         this.defaultSearch = builder.getDefaultSearch();
         this.warnUser = builder.warnUser();
         this.enableDecompositionOfBooleanSum = builder.enableDecompositionOfBooleanSum();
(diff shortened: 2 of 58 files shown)


solver/src/main/java/org/chocosolver/solver/SettingsBuilder.java
```diff
@@ -0,0 +1,947 @@
...
+     *
+     * @param properties the properties to create the `Settings` from
+     * @return the current instance of `SettingsBuilder` with fields set according to the provided properties
+     */
+    public SettingsBuilder fromProperties(Properties properties) {
+        properties.forEach((k, v) -> {
+            String key = k.toString();
+            String value = v.toString();
+            switch (key) {
+                case "modelChecker":
```
jgFages: I would much prefer using constants rather than String values directly

solver/src/main/java/org/chocosolver/solver/SettingsBuilder.java
```diff
@@ -0,0 +1,947 @@
...
+     * Get the predicate to check the solution of the model before running the resolution.
+     * The default value is a predicate that checks if the model is not unsatisfiable.
+     *
+     * @param modelChecker the predicate to check the solution of the model before running the resolution.
+     * @return the current instance
+     */
+    public SettingsBuilder setModelChecker(Predicate<Solver> modelChecker) {
+        this.modelChecker = modelChecker;
+        return this;
+    }
```
jgFages, line 425: FYI, this class could be simplified using Lombok @Builder and @Getter.
This being said if it only for this class, better not introducing a new dependency. 
I am writing this in case you find it relevant for a more general usage.

parsers/src/main/java/org/chocosolver/parser/flatzinc/ast/FConstraint.java
```diff
@@ -408,7 +408,7 @@ public void build(Model model, Datas datas, String id, List<Expression> exps, Li
             IntVar c = exps.get(2).intVarValue(model);
             BoolVar r = exps.get(3).boolVarValue(model);
 
-            if ((boolean) model.getSettings().get("adhocReification").orElse(false)) {
+            if (Boolean.getBoolean(model.getSettings().get("adhocReification").orElse("false"))) {
```
jgFages: as said below, I would prefer pass the map key using a constant ADHOC_REIFICATION rather than the string value directly

ArthurGodet: I agree with the use of constant String

solver/src/test/java/org/chocosolver/solver/variables/impl/AbstractVariableTest.java
```diff
@@ -99,26 +101,19 @@ public void testAddMonitor() throws ContradictionException {
...
-        // Redundancy checker enabled, we expect only 1 m1 and the m2.
-        score.set(0);
-        v1 = model.intVar(0, 20);
-        model.getSettings().setCheckDeclaredMonitors(true);
-        v1.addMonitor(m1);
-        v1.addMonitor(m2);
-        v1.addMonitor(m1);
-        v1.notifyMonitors(IntEventType.VOID);
-        Assert.assertEquals(score.get(), 4);
+        if(!check) {
```
ArthurGodet: To respect the formatting we introduced in the former Pull Request, it should be `if (!check)`

solver/src/test/java/org/chocosolver/solver/variables/impl/AbstractVariableTest.java
```diff
@@ -99,26 +101,19 @@ public void testAddMonitor() throws ContradictionException {
...
+            // Redundancy checker disabled, we expect all monitors to be updated.
+            Assert.assertEquals(score.get(), 5);
+        }else {
```
ArthurGodet: To respect the formatting we introduced in the former Pull Request, it should be `} else {`

solver/src/test/java/org/chocosolver/solver/constraints/nary/sum/IntLinCombTest.java
```diff
@@ -136,7 +136,7 @@ protected Model sum(int[][] domains, int[] coeffs, int b, int op, boolean incr)
     }
 
     protected Model intlincomb(int[][] domains, int[] coeffs, int b, int op, boolean incr) {
-        Model model = new Model(Settings.init().setEnableIncrementalityOnBoolSum(i -> incr));
+        Model model = new Model(SettingsBuilder.init().setThresholdForIncrementalityOnBoolSum(incr?0:Integer.MAX_VALUE));
```
ArthurGodet: I would prefer `incr ? 0 : Integer.MAX_VALUE` for readibility

solver/src/test/java/org/chocosolver/solver/constraints/nary/sum/IntLinCombTest.java
```diff
@@ -115,7 +115,7 @@ public void testNeq() {
 
 
     protected Model sum(int[][] domains, int[] coeffs, int b, int op, boolean incr) {
-        Model model = new Model(Settings.init().setEnableIncrementalityOnBoolSum(i -> incr));
+        Model model = new Model(SettingsBuilder.init().setThresholdForIncrementalityOnBoolSum(incr?0:Integer.MAX_VALUE));
```
ArthurGodet: I would prefer `incr ? 0 : Integer.MAX_VALUE` for readibility

solver/src/main/java/org/chocosolver/solver/Settings.java
```diff
@@ -9,134 +9,111 @@
...
 
-    private boolean lcg = false;
+    private final double ibexContractionRatio;
 
-    private Supplier<IEnvironment> environmentSupplier = () -> new EnvironmentBuilder().fromFlat().build();
+    private final boolean ibexRestoreRounding;
 
-    private Settings() {
-    }
+    private final HashMap<String, String> additionalSettings;
```
ArthurGodet: I'd prefer it to be a `Map` instead of a `HashMap`

solver/src/main/java/org/chocosolver/solver/SettingsBuilder.java
```diff
@@ -0,0 +1,947 @@
...
+
+    private Supplier<IEnvironment> environmentSupplier = () -> new EnvironmentBuilder().fromFlat().build();
+
+    @SuppressWarnings("FieldMayBeFinal") // mutable on purpose, to allow setting values through command line arguments
+    @Option(name = "-P",
+            handler = MapOptionHandler.class,
+            usage = "additional settings that can be set through command line arguments, with the syntax -Dkey=value. " +
+                    "These settings are stored in the additionalSettings map and can be retrieved with the get method. " +
+                    "This allows to set custom settings that are not defined in this class, but that can be used in the code with the get method.")
+    private HashMap<String, String> additionalSettings = new HashMap<>();
```
ArthurGodet: Same here, I'd prefer the type to be a `Map` instead of a `HashMap`

solver/src/main/java/org/chocosolver/solver/SettingsBuilder.java
```diff
@@ -0,0 +1,947 @@
...
+     *
+     * @param defaultMinCardinalityForSumDecomposition cardinality threshold
+     * @return the current instance
+     */
+    public SettingsBuilder setMinCardinalityForSumDecomposition(int defaultMinCardinalityForSumDecomposition) {
+        this.minCardForSumDecomposition = defaultMinCardinalityForSumDecomposition;
+        return this;
+    }
+
+
```
ArthurGodet: There are several two empty lines, where it should only be one for formatting reasons

solver/src/main/java/org/chocosolver/solver/SettingsBuilder.java
```diff
@@ -0,0 +1,947 @@
...
+    public SettingsBuilder setNbMaxLearntClauses(int n) {
+        this.nbMaxLearnt = n;
+        return this;
+    }
+
+    /**
+     * @return <tt>true</tt> if the {@link IntVarLazyLit} propagator uses weak bounds.
+     * @see #setIntVarLazyLitWithWeakBounds(boolean)
+     */
+    public boolean intVarLazyLitWithWeakBounds() {
```
ArthurGodet: By default, I would prefer getters for booleans to start with "is", to ease maintenance and use of the API, and other getters to start by "get"

ArthurGodet (changes requested):
Generally, just some minor changes for formatting reasons.

For the SettingsBuilder class, I would rename most of the getters to be more standard (especially for an API).

And, as @jgFages said, it would be better to rely on constant (`static final`) Strings than use the Strings directly in  the code, to be more robust in case we would like to change the Strings in the future

solver/src/main/java/org/chocosolver/solver/SettingsBuilder.java
jgFages: Agreed

jgFages: Agreed

ArthurGodet (approved):

parsers/src/main/java/org/chocosolver/parser/flatzinc/ast/FConstraint.java
```diff
@@ -2618,6 +2618,8 @@ public void build(Model model, Datas datas, String id, List<Expression> exps, Li
     };
 
 
+    public static final String ADHOC_REIFICATION = "adhocReification";
```
jgFages, line 2621: Is this setting really dedicated to flatzinc ? Shouldn't it belong to the main solver module ?

jgFages (approved):

PR #1184, opened 2026-03-02
add table reformulation for POW + optimized tuples generation

jgFages:


jgFages:
We could also replace

public static Tuples scalar(IntVar[] VARS, final int[] COEFFS, IntVar SCALAR, final int SCALAR_COEFF) {
        Tuples left = generateTuples(TupleValidator.TRUE, true, VARS);
        Tuples tuples = new Tuples(true);
        int n = VARS.length;
        for (int[] tleft : left.tuples) {
            int right = 0;
            for (int i = 0; i < n; i++) {
                right += tleft[i] * COEFFS[i];
            }
            if (right % SCALAR_COEFF == 0 && SCALAR.contains(right / SCALAR_COEFF)) {
                int[] t = new int[n + 1];
                System.arraycopy(tleft, 0, t, 0, n);
                t[n] = right / SCALAR_COEFF;
                tuples.add(t);
            }
        }
        return tuples;
    }
    
    by 
    
    public static Tuples scalar(IntVar resultVar, final int resultCoef, IntVar[] VARS, final int[] COEFFS) {
        return generateTuplesWithResult(
                resultVar,
                values -> IntStream.range(0, values.length).map(i -> values[i] * COEFFS[i]).sum() % resultCoef == 0,
                values -> IntStream.range(0, values.length).map(i -> values[i] * COEFFS[i]).sum() / resultCoef,
                VARS
        );
    }
    
    for code maintenance but performance might be lower so I did not include it

solver/src/main/java/org/chocosolver/solver/constraints/extension/TuplesFactory.java
```diff
@@ -366,12 +366,12 @@ public static Tuples square(IntVar result, IntVar x) {
...
 
     /**
-     * Generate valid tuples for euclidean division constraint: DIVIDEND / DIVISOR = RESULT, rounding towards 0
+     * Generate valid tuples for euclidean division constraint: dividend / divisor = result, rounding towards 0
      *
      * @return a Tuples object, reserved for a table constraint
      */
-    public static Tuples eucl_div(IntVar DIVIDEND, IntVar DIVISOR, IntVar RESULT) {
-        return generateTuples(values -> values[0] / values[1] == values[2], true, DIVIDEND, DIVISOR, RESULT);
+    public static Tuples eucl_div(IntVar result, IntVar dividend, IntVar divisor) {
```
cprudhom, line 373: It is a little bit dangerous to change the order of the parameters, isn't it ?

jgFages: Good remark. The problem is that the result is currently not always in the same position (last argument for modulo, div and times but first position for square) so we would need to make some change anyway if we want to have a clear convention (either all first or all last). I chose the result position at first to allow an IntVar... signature for the other variables (in generateTuplesWithResult and canBeTupledWithResult) :

```
if (TuplesFactory.canBeTupledWithResult(result, base)) {
       return table(new IntVar[]{result, base}, TuplesFactory.square(result, base));
}
```
but we may prefer using an explicit array IntVar[] in which case both positions would be fine (yet a convention either first or last would help avoiding errors). 

Maybe we could simply rename TuplesFactory to make sure the code is reviewed upon solver update ? 

We can also keep existing signatures and be cautious when calling generateTuplesWithResult.

Let me know what you prefer or if you have other suggestions

cprudhom (approved):

PR #1185, opened 2026-03-27
Enhance MiniSat and LCG with improved reason management and optimizations

cprudhom:
This appears to be a big PR because all explained propagators have been modified but the main modifications are in the `sat` package.
The way reasons are managed has been deeply modified to reduce GC by default (n-ary reasons are recycled) even if it is still manageable to use the previous version.
In addition, Literal Block Distance has been implemented and is used when the clause store needs to be reduced to better select the ones to keep.
Also, the way tasks involved in the profile of  the cumulative constraint are stored has changed too are (when LCG is activated).
+ some other corrections.


solver/src/main/java/org/chocosolver/solver/variables/Task.java
```diff
@@ -392,57 +392,60 @@ public void propagate(int evtmask) throws ContradictionException {
...
+                        hasFiltered = updateEst(end.getLB() - duration.getUB(), this);
+                        hasFiltered |= updateLst(end.getUB() - duration.getLB(), this);
+
+                        hasFiltered |= updateEct(start.getLB() + duration.getLB(), this);
+                        hasFiltered |= updateLct(start.getUB() + duration.getUB(), this);
+
+                        hasFiltered |= updateMinDuration(end.getLB() - start.getUB(), this);
+                        hasFiltered |= updateMaxDuration(end.getUB() - start.getLB(), this);
+                    }
+                    // TODO : add filtering for enumerated domains
```
jgFages, line 443: TODO still relevant?

jgFages (approved):

solver/src/main/java/org/chocosolver/sat/IReasonManager.java
```diff
@@ -0,0 +1,622 @@
...
+            }
+            reason.setUp(ci, this);
+            currentReason.set(cr + 1);
+            return reason;
+        }
+
+        private void ensureCapacity(int startIndex, int size) {
+            while (startIndex + size > buffer.length) {
+                buffer = Arrays.copyOf(buffer, (int) (buffer.length * 1.5));
+            }
```
ArthurGodet: It might be a small optimization, but I think it would be better to have a while loop to determine the new needed size for buffer array, and only then call `Arrays.copyOf()`. Most of the time, the current while loop will probably be called only once, but I think this small optimization might be useful in a few cases, with the same cost if call once.

solver/src/test/java/org/chocosolver/sat/IReasonManagerTest.java
ArthurGodet, line 1: `System.out.println()` should be removed

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/PropagatorCumulative.java
```diff
@@ -474,20 +483,33 @@ private boolean scalableTimeTableFilterLct(final Task task, final IntVar height)
...
+                                break;
+                            }
+                            int next_point = Math.max(profile.getStartRectangle(j), task.getLst());
+                            if (next_point < begin) {
+                                begin = next_point; // next time point for pointwise explanation
+                            } else
+                                // the lst did not change upon lct filtering.
+                                // This can happen when lb(duration) = 0
+                                break;
+                        } while (begin > profile.getStartRectangle(j));
```
ArthurGodet, line 509: I don't understand the purpose of this do-while loop

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/PropagatorCumulative.java
```diff
@@ -439,20 +435,33 @@ private boolean scalableTimeTableFilterEst(final Task task, final IntVar height)
...
+                                break;
+                            }
+                            int next_end = Math.min(task.getEct(), profile.getEndRectangle(j));
+                            if (next_end > end) {
+                                end = next_end; // next time point for pointwise explanation
+                            } else
+                                // the ect did not change upon est filtering.
+                                // This can happen when lb(duration) = 0
+                                break;
+                        } while (end < profile.getEndRectangle(j));
```
ArthurGodet, line 461: I don't understand the purpose of this do-while loop

ArthurGodet (changes requested):

solver/src/main/java/org/chocosolver/sat/IReasonManager.java
jgFages: or even without while ?
```
if (startIndex + size > buffer.length) {
   buffer = Arrays.copyOf(buffer, (int) Math.max(startIndex + size, buffer.length * 1.5));
}
```

solver/src/main/java/org/chocosolver/solver/variables/Task.java
cprudhom: I didn't add the TODO in the first place, but I think that achieving AC on Task when domains are enumerated can really be costly. If that's a need, I suggest the user to explicitly add a constraint. @ArthurGodet what do you think ?

solver/src/main/java/org/chocosolver/solver/constraints/nary/cumulative/PropagatorCumulative.java
cprudhom: This is due to the pointwise explanation.
When lcg is on, the pointwise explanation is stronger and more reusable than big-step explanation (ie, the previous version).
"For a profile with length greater than the duration of task j we may need to pick more than one time point in a profile" ("Explaining the cumulative propagator?", Schutt et al.  https://people.eng.unimelb.edu.au/pstuckey/papers/cumulative.pdf)
The details of this choice can be found p. 12 and 15.

ArthurGodet: Ok, I see

solver/src/main/java/org/chocosolver/solver/variables/Task.java
ArthurGodet: I agree on removing this TODO, as filtering algorithms on Tasks (the ones for Cumulative and Disjunctive constraints) always filter on bounds (and never on values inside the domain), and Tasks are designed to be used for these two constraints.

ArthurGodet (approved):

PR #1188, opened 2026-04-03
add arithm preprocessing to detect trivial cases

jgFages:


ArthurGodet (approved):

ArthurGodet:
I agree on the proposed preprocessing. But why are some tests failing due to these changes ?

jgFages:
> I agree on the proposed preprocessing. But why are some tests failing due to these changes ?

Ah, those optimization are no longer possible within a dynamic context...
So even the already existing prepocessing below was wrong actually : 

```
default Constraint arithm(IntVar var1, String op, IntVar var2) {
            if (var2.isInstantiated()) {
                return arithm(var1, op, var2.getValue());
            }
            if (var1.isInstantiated()) {
                return arithm(var2, Operator.getFlip(op), var1.getValue());
            }
...
```

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
```diff
@@ -100,15 +100,52 @@ public interface IIntConstraintFactory extends ISelf<Model> {
...
-     * @param var  a variable
+     * @param x  a variable
      * @param op   an operator
      * @param cste a constant
      */
-    default Constraint arithm(IntVar var, String op, int cste) {
-        return new Arithmetic(var, Operator.get(op), cste);
+    default Constraint arithm(IntVar x, String op, int cste) {
+        // this preprocessing is no longer permitted within a dynamical context
+        if (ref().getSolver().getNodeCount() == 0) {
```
cprudhom: I used to use `solver.getSearchState() == SearchState.NEW`  for instance in `cstr.impliedBy(bvar)`.
I would suggest either to call this method or to create a new method with an accurate name that indicates that such preprocessing can be done safely.

cprudhom:
Wouldn't be better to move this simplification to `IntConstraintFactory` ?
That's where preprocessing are commonly applied.

jgFages:
> Wouldn't be better to move this simplification to `IntConstraintFactory` ? That's where preprocessing are commonly applied.

It is already in `IIntConstraintFactory`, perhaps you meant `PreProcessing.java` ?

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
cprudhom: Well, there exists a method like that: `solver.isSolving()`

PR #1195, opened 2026-04-23
Jg/element

jgFages:


solver/src/main/java/org/chocosolver/solver/constraints/binary/element/PropElementIn.java
```diff
@@ -0,0 +1,268 @@
...
+			firstPropagation();
+		} else {
+			nextPropagation();
+		}
+	}
+
+	@Override
+	public void propagate(int idxVarInProp, int mask) throws ContradictionException {
+		forcePropagate(PropagatorEventType.CUSTOM_PROPAGATION);
+	}
```
ArthurGodet, line 126: Is it useful to make the propagator react on fine events if the fine propagate() function calls the "default" propagate() function ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/PropElementIn.java
```diff
@@ -0,0 +1,268 @@
...
+	private void passivate() {
+		if (index.isInstantiated() && !result.isInstantiated() && result.hasEnumeratedDomain()) {
+			setPassive();
+		}
+	}
+
+	private Integer getSupport(IntIterableSet set, Integer watcher) {
+		if (watcher != null && result.contains(watcher)) {
+			return watcher;
+		}
```
ArthurGodet, line 189: Since `watcher` is a value from `watchLiterals` array, which is an `int[]`, I think it would be best to make the variable `watcher` of this function to be an `int` instead of an `Integer` (to avoid the cost of autoboxing), since `watcher` can't be `null`.

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/PropElementIn.java
```diff
@@ -0,0 +1,268 @@
...
+	public ESat isEntailed() {
+		if (index.getUB() < 0 || index.getLB() >= values.length) {
+			return ESat.FALSE;
+		}
+		if (isCompletelyInstantiated()) {
+			IntIterableSet set = values[index.getValue()];
+			int value = result.getValue();
+			if (!set.contains(value)) {
+				throw new UnsupportedOperationException();
+			}
```
ArthurGodet: Should we keep this exception ? Isn't it the purpose of the following line (`ESat.eval(set.contains(value));`) to evaluate this ?

ArthurGodet (changes requested):

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/PropElementIn.java
jgFages: yes the algorithm is then a bit different 

```
		if (PropagatorEventType.isFullPropagation(evtmask)) {
			firstPropagation();
		} else {
			nextPropagation();
		}
```

In the firstPropagation, we compute the feasible values union and call removeAllValuesBut
In the latter, as sets are disjoint, we remove incrementally values as indexes are removed, without recomputing the feasible union nor calling removeAllValuesBut (that is a bit time consuming)

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/PropElementIn.java
cprudhom: Watcher can be null (l.217 or l.238) but I suppose we can use negative value to catch null cases.

solver/src/main/java/org/chocosolver/solver/variables/IResultVariableFactory.java
```diff
@@ -157,22 +158,55 @@ default IntVar count(String name, IntVar value, IntVar... vars) {
...
+	 * @return an integer variable
 	 */
 	default IntVar element(String name, int[] table, IntVar index, int offset) {
-		IntVar result = ref().intVar(name, index.stream()
-				.filter(v -> v >= offset && v < table.length + offset)
-				.map(v -> table[v - offset]).toArray());
-		if (!result.isInstantiated()) {
-			ref().element(result, table, index, offset).post();
-		} else if (index.getLB() < offset || index.getUB() > offset + table.length - 1) {
+		int n = table.length;
```
cprudhom, line 182: Don't want to deport (or use) the logic from IIntConstraintFactory.element(...) ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/ElementFactory.java
```diff
@@ -6,71 +6,144 @@
...
-        i++;
-        while (i < TABLE.length - 1) {
-            if (up && TABLE[i] > TABLE[i + 1]) {
-                c++;
-                up = false;
-            } else if (!up && TABLE[i] < TABLE[i + 1]) {
-                c++;
-                up = true;
+        // large tables
+        if (n > 50 && index.hasEnumeratedDomain()) {
```
cprudhom, line 55: Do you have any intuition about this "magic" number ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/ElementFactory.java
```diff
@@ -6,71 +6,144 @@
...
+            // table reduction by merging same value cells
+            if (nbValues * 10 < n && value.hasEnumeratedDomain() && index.hasEnumeratedDomain()) {
```
cprudhom, line 57: Same question here

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/ElementFactory.java
jgFages: I guess we could make a field in settings for it ?

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/PropElementIn.java
ArthurGodet: Yes, I understand, but isn't it done without declaring fine propagation ? My comment is a bit of cherry picking but I think the code would be clearer if using only the "non-fine events" propagate() if it is not necessary

solver/src/main/java/org/chocosolver/solver/constraints/binary/element/PropElementIn.java
ArthurGodet: My earlier comment was on code that would later be changed (and in the new code, watcher could be null)

PR #1198, opened 2026-05-01
Improve filtering on signs of variables in PropDivXYZ

ArthurGodet:


ArthurGodet:
Test failing (SparseBitSetTest.testBacktracking) seems to be quite random : it does happen every time in local due to randomness.
That said, it is not up to the modifications in this PR.

jgFages (approved):

cprudhom (approved):

PR #1199, opened 2026-05-03
Fix SparseBitSet.prevClearBit()

ArthurGodet:


cprudhom:
I have two remarks/questions:
1. does `nextClearBit()` need to be patched too?
2. can you provide an example where such situation appears?

ArthurGodet:
> I have two remarks/questions:
> 
>     1. does `nextClearBit()` need to be patched too?
> 
>     2. can you provide an example where such situation appears?

1. No, it already had a similar patch as the one I did. I just uniformised code between the two functions.
2. Yes, I added `testPrevClearBit3()` in `SparseBitSetTest` : before the correction, it returned 895 instead of 893.

solver/src/test/java/org/chocosolver/memory/structure/SparseBitSetTest.java
```diff
@@ -153,6 +153,18 @@ public void testPrevClearBit2() {
...
 
+  @Test
+  public void testPrevClearBit3() {
+    IEnvironment env = new EnvironmentTrailing();
+    IStateBitSet bs = env.makeSparseBitset(64);
+    bs.set(14);
+    bs.set(20);
+    bs.set(899);
+    bs.set(929);
+
```
cprudhom, line 164: I would add the following assertion: 
```java
Assert.assertEquals(bs.prevClearBit(60), 60);
Assert.assertEquals(bs.prevClearBit(300), 300);
``` 
just to make sure it works on various cases.

cprudhom (approved):

cprudhom (approved):

PR #1200, opened 2026-05-05
remove slf4j-nop dependency to avoid conflicts in applications using …

jgFages:
…choco-solver


cprudhom:
Could this be only activated to test ?
```xml
        <dependency>
            <groupId>org.slf4j</groupId>
            <artifactId>slf4j-nop</artifactId>
            <version>${slf4j.nop.version}</version>
            <scope>test</scope>
        </dependency>
```
?

Issue #1207, opened 2026-06-04
Move seed parameter from Model and RegParser to Settings and SettingsBuilder


YoanThomasPhd:
Currently, it is not possible (to my knowledge) to configure the model seed directly from a parser.

I propose moving the seed parameter from `Model` and `RegParser` to `Settings` and `SettingsBuilder` so that it can be configured through parser options and centralized alongside other solver settings.

A minimal implementation could look like:

`Settings.java`
```java

protected long seed;

public void setSeed(long seed) {
    this.seed = seed;
}

public long getSeed() {
    return seed;
}

```
`SettingsBuilder.java`
``` java

public static final String SEED = "seed";

@Option(name = "-seed", usage = "Set the seed for the random number generator.")
protected long seed = 1_000_000_007L;

```
`Model.java`
``` java

public void setSeed(long seed) {
    settings.setSeed(seed);
}

public long getSeed() {
    return settings.getSeed();
}
```

cprudhom:
Ok for me. Can you create a PR with the modification?

cprudhom:
I suppose this can be related to #757

PR #1209, opened 2026-06-05
Request #1207

YoanThomasPhd:
Move seed from Model and RegParser to Settings and SettingsBuilder.

The seed is now accessible via model.getSettings().getSeed(). After initialization, the seed can no longer be changed.

### DIFF
--- a/parsers/src/main/java/org/chocosolver/parser/RegParser.java
+++ b/parsers/src/main/java/org/chocosolver/parser/RegParser.java
@@ -159,9 +159,6 @@ public abstract class RegParser extends SettingsBuilder implements IParser {
             "--nb-cores"}, usage = "Number of cores available for parallel search (default: 1).")
     protected int nb_cores = 1;
 
-    @Option(name = "-seed", usage = "Set the seed for random number generator. ")
-    protected long seed = 1_000_000_007L;
-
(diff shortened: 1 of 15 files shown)


ArthurGodet (approved):

ArthurGodet (changes requested):
The modifications of code are fine to me. However, since some default values have been changed, some expected values need to be change in tests

cprudhom:
Since `dom/wdeg` is the default strategy, changing the seed affects a large number of tests.
Before approving the PR, we need to make sure that the differences are indeed caused solely by this (which may take a little time).

YoanThomasPhd:
I won't have time to check and fix the tests for now, I will try to verify the validity of `dom/wdeg` and the others later this summer
For now I will assume that everything is valid and use this fix for my tests with different seed!

cprudhom:
I’ve taken a closer look at the PR. I think we can simplify it a little; this will affect fewer classes and retain the spirit of the PR.
Essentially, I propose keeping `Model.getSeed()`, removing `Model.setSeed(long)` and setting the seed when the model is created by retrieving the value declared in `Settings`.

If that’s OK with you, I can take care of making the necessary changes.

PR #1211, opened 2026-06-08
fix: avoid spurious UNSAT under LCG on reasonless (C_Fail) conflicts

### LINKED ISSUE #1210: [BUG] LCG reports UNSAT on a satisfiable model when the value selector picks a pruned value
**Describe the bug**

With lazy clause generation enabled, the solver can report UNSAT on a satisfiable model. It
happens when a value selector instantiates a variable to a value no longer in its domain.
IntDomainMiddle does this readily, since it returns min + (max - min) / 2 without checking the
value is still present. That enqueues an already-false literal with no reason (a C_Fail
conflict), and conflict analysis reads its level off the {0, 0} sentinel clause, gets 0, and
backjumps above the root, concluding UNSAT.

**To Reproduce**

Run this:

    Model model = new Model(new SettingsBuilder().setLCG(true).build());
    IntVar x = model.intVar("x", 0, 4);
    model.arithm(x, "!=", 2).post();          // remove the midpoint value
    solver.setSearch(Search.intVarSearch(
            new InputOrder<>(model), new IntDomainMiddle(true),
            DecisionOperatorFactory.makeIntEq(), x));
    System.out.println(solver.solve());        // prints false

A larger, real-world trigger (a de Bruijn sequence under first_fail / indomain_split) is in
the minimal repro repository: https://github.com/rasros/choco-lcg-false-unsat

**Expected behavior**

The model is satisfiable (x can be 0, 1, 3 or 4), so the solver should never report UNSAT.
Disabling LCG, or using the default search, correctly returns SAT.

**Possible solution**

I have a PR ready.

**Environment (please complete the following information):**

- Choco-solver version: 6.0.1 (also reproduces on master)
- JRE: 21

rasros:
Fixes #1210

Changes proposed in this PR:

- Fix a spurious UNSAT under LCG: when a value selector instantiates a variable to a value
  already pruned from its domain (IntDomainMiddle does this on domains with a hole near the
  middle), the reasonless C_Fail conflict had its level read off the {0, 0} sentinel clause,
  so analyze() got 0 and backjumped above the root. It now backs up one level (clamped to the
  root) to refute the offending decision instead.
- Add a regression test, LCGTest.testSpuriousUnsatOnPrunedValueDecision, in the {1s,lcg} group.
- Add myself to the authors of MiniSat.java.

@chocoteam/core-developer

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/values/IntDomainMiddle.java
+++ b/solver/src/main/java/org/chocosolver/solver/search/strategy/selectors/values/IntDomainMiddle.java
@@ -22,6 +22,7 @@ import java.util.function.ToDoubleFunction;
  * would result in no inference)
  *
  * @author Charles Prud'homme, Jean-Guillaume Fages
+ * @author Rasmus Ros
  * @since 2 juil. 2010
  */
 public class IntDomainMiddle implements IntValueSelector {
@@ -61,12 +62,16 @@ public class IntDomainMiddle implements IntValueSelector {
             } else {
                 value = min + (max - min + 1) / 2;
             }
+            // the midpoint may have been pruned from the domain; snap to an existing value so the
+            // generated decision is consistent (an out-of-domain value leads to a spurious failure)
+            value = var.nextValue(value - 1);
         } else {
             if (roundingPolicy == FLOOR) {
                 value = var.getLB();
             } else {
                 value = var.getUB();
             }
+            value = var.previousValue(value + 1);
         }
         return value;
     }
--- a/solver/src/test/java/org/chocosolver/solver/lcg/LCGTest.java
+++ b/solver/src/test/java/org/chocosolver/solver/lcg/LCGTest.java
@@ -13,6 +13,9 @@ import org.chocosolver.solver.Solver;
 import org.chocosolver.solver.constraints.Constraint;
 import org.chocosolver.solver.constraints.nary.cnf.LogOp;
 import org.chocosolver.solver.search.strategy.Search;
+import org.chocosolver.solver.search.strategy.assignments.DecisionOperatorFactory;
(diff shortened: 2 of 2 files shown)


cprudhom:
Hi,

Thank you for the PR and your contribution.
However, I think that the problem comes from the value selector (`IntDomainMiddle`) that should provide a value on the basis of which a correct decision must be made; namely, a decision that at least reduces the domain a variable whilst remaining consistent. Indeed, generating decisions that are trivially inconsistent does not help to advance the exploration of the search space.

A correction could be:

```java
    public int selectValue(IntVar var) {
        int value;
        if (var.hasEnumeratedDomain()) {
            int min = var.getLB();
            int max = var.getUB();
            if (roundingPolicy == FLOOR) {
                value = min + (max - min) / 2;
            } else {
                value = min + (max - min + 1) / 2;
            }
            value = var.nextValue(value - 1);
        } else {
            if (roundingPolicy == FLOOR) {
                value = var.getLB();
            } else {
                value = var.getUB();
            }
            value = var.previousValue(value + 1);
        }
        return value;
    }
```

rasros:
That's fair. I'll have a look tonight if it's something that touches more value selectors or not. If not I'll have a go at it.

rasros:
So no other value selector can produce out-of-domain values. But it seems like the case for IntDomainMiddle to do it is legit: it's normally paired with split decisions, where the midpoint is just a bound and doesn't need to be in the domain. The existing IntDomainMiddleTest actually relies on that, so snapping it breaks those tests. The value only needs to be in the domain when it's used with an = decision, like my repro does, and the selector has no way to know that. So I don't think the selector is the right place to fix it.


I'll close this PR since the fix doesn't belong in the selector. But the bug itself is real and reachable through MiniZinc's indomain_middle (the parser maps it to IntDomainMiddle + =), so it's worth fixing at the conflict-analysis level.

PR #1218, opened 2026-08-14
Fix issue #1214

### LINKED ISSUE #1214: [BUG] PropTimesNaive and PropTimesNaiveWithLong perform unsound pruning due to float rounding in div()
## Description

The bound propagation of `X * Y = Z` uses `float` arithmetic to compute quotient bounds in `div()` (case 5). `float` has a 24-bit mantissa, so not every integer above 2^24 ≈ 16.7M is exactly representable. When a bound of `Z` exceeds that range and is not exactly representable as a `float`, an exact integer quotient can be computed as `q - epsilon`, its floor becomes `q - 1`, and a supported value is pruned. The propagation is therefore unsound: a feasible instantiation is reported infeasible, and in optimization valid (even Pareto-optimal) solutions are silently removed, with no warning or exception.

`PropTimesNaiveWithLong` is equally affected, because its `div()` method still performs the quotient-bound computations using `float`.

## Minimal reproducible example

All variable bounds, including `z`'s, are deliberately kept below `IntVar.MAX_INT_BOUND` (= `Integer.MAX_VALUE / 100` = 21_474_836):

```java
Model m = new Model();
IntVar x = m.intVar("x", 0, 7_625);
IntVar y = m.intVar("y", 7_335, 13_365);
IntVar z = m.intVar("z", 0, 20_381_625); // below IntVar.MAX_INT_BOUND
m.times(x, y, z).post();
m.arithm(x, "=", 1_525).post();
m.arithm(y, "=", 13_365).post();
System.out.println(m.getSolver().solve());
// Actual:   false
// Expected: true, since 1_525 * 13_365 = 20_381_625, which lies in [0, 20_381_625]
```

The same result is obtained when posting `new Constraint("T", new PropTimesNaiveWithLong(x, y, z))` instead of `m.times(...)`.

Note for a regression test: the failure requires the domains to shrink *during* propagation, as above (equalities posted as constraints on non-singleton domains). A variant where `x`, `y`, `z` are created as constants (`m.intVar("x", 1_525)`, etc.) does **not** trigger the bug, because the fatal division never runs on the intermediate bounds — such a test would pass on the current code.

## Root cause (propagation trace)

1. `x = 1525` propagates; `mul()` tightens `z` to `[1525*7335, 1525*13365] = [11_185_875, 20_381_625]`.
2. `y = 13365` propagates; the propagator recomputes `x`'s bounds via `div(x, z.LB, z.UB, y.LB, y.UB)`, i.e. case 5 with `b = 20_381_625`, `c = d = 13_365`.
3. `(float) 20_381_625` is not exactly representable (odd value above 2^24, ulp = 2); it rounds to `20_381_624.0f`.
4. `20_381_624f / 13_365 = 1524.9999f` (the exact quotient is 1525.0).
5. `max = (int) Math.round(Math.floor(1524.9999f)) = 1524` → `x.updateUpperBound(1524)` → contradiction with `x = 1525`.

Since every bound in the example is below `IntVar.MAX_INT_BOUND`, the failure cannot be attributed to using values outside Choco's documented recommended range. More generally, silently reporting infeasibility for a valid instantiation is an unsound result: if some arithmetic range were unsupported, the model should be rejected explicitly rather than supported values being silently pruned.

## Real-world impact

We hit this in a covering/nesting model (widths × cut length = area, dimensions in mm) with `Solver.findParetoFront` over three objectives. One genuinely Pareto-optimal solution was silently absent from the returned front (5 points instead of 6); brute-force enumeration confirmed the missing point is feasible and non-dominated. The pruning happens deep inside propagation, so nothing signals that the result is incomplete.

## Affected versions

Reproduced with 4.10.18 and 6.0.1: the minimal example above returns `false` on both versions. Solver statistics after `solve()` show `Nodes: 0` and `Fails: 1`, confirming that the contradiction is raised during root propagation, before any branching occurs. The solver reports `Complete search - No solution`, thereby declaring a feasible model infeasible.

At the time of writing, the `float` divisions are present in commit `f210dc0`:
- https://github.com/chocoteam/choco-solver/blob/f210dc0aac61ceb0f8dc38d31d49cdbe785276c8/solver/src/main/java/org/chocosolver/solver/constraints/ternary/PropTimesNaive.java#L90-L93
- https://github.com/chocoteam/choco-solver/blob/f210dc0aac61ceb0f8dc38d31d49cdbe785276c8/solver/src/main/java/org/chocosolver/solver/constraints/ternary/PropTimesNaiveWithLong.java#L83-L86

## Suggested fix

The robust fix is maybe to compute the signed floor and ceiling quotient bounds directly with exact integer arithmetic on `long` (`Math.floorDiv` and a ceiling counterpart), avoiding floating-point rounding entirely. This is especially natural for `PropTimesNaiveWithLong`, whose intermediate values are meant to be `long` already.

A minimal alternative would be to replace `float` with `double` in `div()`. Since the operands are `int`-valued bounds, this should provide sufficient precision for these quotient-bound calculations. Nevertheless, exact integer floor and ceiling division on `long` is preferable because it removes floating-point reasoning from the propagator entirely.

## Additional independent issue in `PropTimesNaiveWithLong.mul()`

Separately from the precision bug above: in `mul()`, the products are computed in `int` before being widened to `long` (`long min = Math.min(a * c, ...)`), so overflow can occur before the `long` protection applies; casting one operand (`(long) a * c`) fixes it. I can open a distinct ticket for this one if preferred.

---

Happy to provide the full Pareto model or additional traces if useful.

cprudhom:
See #1214 for details

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/variables/IntVar.java
+++ b/solver/src/main/java/org/chocosolver/solver/variables/IntVar.java
@@ -428,7 +428,7 @@ public interface IntVar extends ICause, Variable, ArExpression {
      * @throws ContradictionException if the domain become empty due to this action
      */
     default boolean updateLowerBound(long value, ICause cause) throws ContradictionException {
-        return updateUpperBound(value, cause, cause.defaultReason(this));
+        return updateLowerBound(value, cause, cause.defaultReason(this));
     }
 
     /**
--- a/solver/src/main/java/org/chocosolver/solver/variables/IntVar.java
+++ b/solver/src/main/java/org/chocosolver/solver/variables/IntVar.java
@@ -428,14 +428,41 @@ public interface IntVar extends ICause, Variable, ArExpression {
      * @throws ContradictionException if the domain become empty due to this action
      */
     default boolean updateLowerBound(long value, ICause cause) throws ContradictionException {
+        return updateUpperBound(value, cause, cause.defaultReason(this));
+    }
+
+    /**
+     * Updates the lower bound of the domain of <code>this</code> to <code>value</code>.
+     * The instruction comes from <code>propagator</code>.
+     * <p>
+     * This method deals with <code>value</code> as <b>long</b>.
+     * If such a long can be safely cast to an int, this falls back to regular case (int).
+     * Otherwise, it can either trivially do nothing or fail.
+     * </p>
+     * <ul>
+     * <li>If <code>value</code> is smaller than the lower bound of the domain, nothing is done and the return value is <code>false</code>,</li>
+     * <li>if updating the lower bound to <code>value</code> leads to a dead-end (domain wipe-out),
+     * a <code>ContradictionException</code> is thrown,</li>
+     * <li>otherwise, if updating the lower bound to <code>value</code> can be done safely,
+     * the event type is created (the original event can be promoted) and observers are notified
+     * and the return value is <code>true</code></li>
+     * </ul>
+     *
+     * @param value new lower bound (included)
+     * @param cause updating releaser
+     * @param reason the reason why the lower bound is updated
+     * @return true if the lower bound has been updated, false otherwise
+     * @throws ContradictionException if the domain become empty due to this action
+     */
+    default boolean updateLowerBound(long value, ICause cause, Reason reason) throws ContradictionException {
         if ((int) value != value) { // cannot be cast to an int
             if (value < getLB()) {
                 return false;
             } else { // then value >> getLB, this fails
-                return updateLowerBound(getUB() + 1, cause);
+                return updateLowerBound(getUB() + 1, cause, reason);
             }
         } else {
-            return updateLowerBound((int) value, cause);
+            return updateLowerBound((int) value, cause, reason);
         }
     }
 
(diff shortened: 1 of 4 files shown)


ArthurGodet (approved):

JulienBranco:
Thanks a lot for the fix.

While reading the diff I noticed something in "IntVar.java" that I think may be an unintended typo, though I may well be misreading it:

```
default boolean updateLowerBound(long value, ICause cause) throws ContradictionException {
    return updateUpperBound(value, cause, cause.defaultReason(this));
}
```

The two-argument "long" overload seems to delegate to "updateUpperBound" rather than to the new three-argument "updateLowerBound". The "updateUpperBound(long, ICause)" counterpart just below delegates correctly, which is what made me wonder whether this one was a copy-paste slip.

If that reading is right, it could be easy to miss in CI: as far as I can tell, the internal caller in "PropTimesNaiveWithLong" now goes through the three-argument version, so the new regression tests for #1214 do not appear to exercise this overload. An "int" argument would also resolve to the "int" overload. It would still be reachable from user code, since it's a "default" method on the public "IntVar" interface.

Sorry if I've overlooked something wrong here — and thanks again for looking into the original issue.

ArthurGodet:
You are right, it should be delegating the call to updateLowerBound and not to updateUpperBound

solver/src/main/java/org/chocosolver/solver/variables/IntVar.java
```diff
@@ -428,14 +428,41 @@ default boolean updateLowerBound(int value, ICause cause) throws ContradictionEx
      * @throws ContradictionException if the domain become empty due to this action
      */
     default boolean updateLowerBound(long value, ICause cause) throws ContradictionException {
+        return updateUpperBound(value, cause, cause.defaultReason(this));
```
ArthurGodet: It should be updateLowerBound here instead of updateUpperBound

ArthurGodet (changes requested):

jgFages (approved):

Issue #1175, opened 2025-11-12
[BUG] Nested calls to `ISetIterator` go to unexpected behavior


cprudhom:
This is related to #1174 
Indeed, `ISetIterator` are reusable, to prevent from creating iterators anytime an iteration is needed.
But it may happen that the same iterator is used inside an iteration (ex: iteration to filter and iteration when instantiated). In that case, the top iteration may not loop over all the values.

A possible fix would be to make ISetIterator extends `AutoCloseable` too, to catch end of usage in combination of PoolManager. That way, it could be possible to state if an iterator is still in used (with `AutoCloseable`); if so, a new one would be created or an old one would be recycled; if not, it would be push as recyclable.

cprudhom:
```java
@Test(groups="1s", timeOut=5000)                        
public void testSetIterator(){                         
    ISet set = new Set_BitSet(0);                      
    //values = {2,3,5,6,11}                            
    set.add(2);                                        
    set.add(3);                                        
    set.add(5);                                        
    set.add(6);                                        
    set.add(11);                                       
                                                       
    for(int v : set){                                  
        try {                                          
            for (int w : set) {                        
                System.out.printf("(%d,%d)\n", v, w);  
                if (v == 3 && w == 5) {                
                    throw new ContradictionException();
                }                                      
                if (v == 5 && w == 6) {                
                    break;                             
                }                                      
            }                                          
        }catch (ContradictionException ce){            
                                                       
        }                                              
    }                                                  
}                                                      
```
Should print:
```
(2,2)
(2,3)
(2,5)
(2,6)
(2,11)
(3,2)
(3,3)
(3,5)
(5,2)
(5,3)
(5,5)
(5,6)
(6,2)
(6,3)
(6,5)
(6,6)
(6,11)
(11,2)
(11,3)
(11,5)
(11,6)
(11,11)
````
but prints:
````
(2,2)
(2,3)
(2,5)
(2,6)
(2,11)
````

cprudhom:
Same behavior with:
```java
    @Test(groups = "1s", timeOut = 5000)
    public void testIntVarValueIterator() {
        Model m = new Model();
        IntVar var = m.intVar("v", new int[]{2,3,5,6,11});

        for (int v : var) {
            try {
                for (int w : var) {
                    System.out.printf("(%d,%d)\n", v, w);
                    if (v == 3 && w == 5) {
                        throw new ContradictionException();
                    }
                    if (v == 5 && w == 6) {
[...]

PR #1220, opened 2026-08-15
Fix issue #1175

ArthurGodet:
The changes in this Pull Request modify the implementations of `ISetIterator`.

On the one hand, it allows two things : 
- Imbricated for-each loop work correctly for `ISetIterator`
- `ISetIterator` now respects the contract of `Iterator` (which is the independency of iterators on a given `Iterable` object)

On the other hand, it changes the fundamental philosophy of reusable iterators (not relying too much on the Garbage Collector).

That being said, I didn't find any existing solution in Java to know if an `Iterator` is being used, which makes impossible to solve the problem of imbricated for-each loop with reusable iterators. Therefore, we have two possibilities : 
- Changing the implementations as done in this PR
- Refusing this PR and well document the fact that for-each loop would not work correctly if imbricated for ISetIterator (or that `newIterator()` function should be specifically used instead, but it breaks the ease of visibility offered by for-each loops)

Let note that many `ISetIterator` implementations rely on a unique integer (basically an index), which should not give too much work to the GC. Furthermore, it is very similar to the implementations of `Iterator<T>` of `List<T>` in the JDK (at least the implementation Eclipse Temurin), and therefore we can expect it is a hehaviour for which the JVM is well-trained/implemented and we should not have too much impact on performance (I found none significant on small examples, but it might be interesting to monitor performance on a diverse family of problems and instance sizes (such as XCSP or MiniZinc) to confirm it).

### DIFF
--- a/solver/src/main/java/org/chocosolver/util/objects/setDataStructures/bitset/Set_Std_BitSet.java
+++ b/solver/src/main/java/org/chocosolver/util/objects/setDataStructures/bitset/Set_Std_BitSet.java
@@ -29,7 +29,6 @@ public class Set_Std_BitSet extends AbstractSet implements ISet.WithOffset {
 	private final IStateInt card;	// enables to get the cardinality in O(1)
 	private final int offset;		// allow using negative numbers
 	private final S64BitSet values;
-	private final ISetIterator iter = newIterator();
 
 	//***********************************************************************************
 	// CONSTRUCTOR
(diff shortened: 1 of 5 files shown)


ArthurGodet:
As discussed with @jgFages, the question seems to have been solved by avoiding imbricated for-each loops altogether within the code of the solver, and that newIterator() should be called whenever one wants to use imbricated for-each loops (but there are only few cases when the same iterator would be use (such as within a propagator and the search heuristic for example)).

cprudhom:
In fact, this work was carried out at the start of the year and incorporated into release [5.0.0](https://github.com/chocoteam/choco-solver/releases/tag/v5.0.0).
It should therefore be regarded as one of the best practices that people should be aware of and, consequently, document.
Especially since, even though the implementation is similar to `Iterator<T> of List<T>` in the JDK, it could also be argued that the way these iterators are used during propagation bears little resemblance to what can be done with a list in a conventional context.

PR #1234, opened 2026-08-21
LCG update

cprudhom:
Some improvements to the LCG framework:
- MiniSat now manages new WL in a circular way (like depicted in "Optimal Implementation of Watched Literals and More General Techniques", Ian P. Gent, JAIR 2013)
- When a clause is learned, the 2nd watched-literal is the one with the highest decision level (as defined in "An Extensible SAT-solver", Niklas Eén, Niklas Sörensson)
- LCG now call for DB reduction before adding the last nogood
- Profie (for Cumulative) lazily computes the set of tasks involved in a conflict


ArthurGodet (approved):

ArthurGodet (changes requested):
The code changes seem fine to me. The only remaining thing to dois change the number of expected nodes in tests in choco-parsers module.

Maybe we should consider removing these specific asserts (and keep only the ones of the objective value or number of solutions). What do you think ? @jgFages an opinion on the matter ?

cprudhom:
In my opinion, we should keep these metrics. It is admittedly a bit of a pain to update them whenever we make a change that we know will affect them, but it also helps us spot regressions caused by other, indirect changes (such as, for example, when we change certain constraints).

ArthurGodet (approved):

jgFages:
I would tend to agree with both :-) 
It is good to keep (at least some of) them in order to detect potential regressions.
It is also acceptable to remove some node/fail counts, especially when using dynamic strategies.

jgFages (approved):

PR #1236, opened 2026-08-21
Fix 1233

### LINKED ISSUE #1233: [BUG] Wrong entailment in PropAbsolute when the result is zero causes missing solutions and false UNSAT
## Describe the bug

`PropAbsolute.isEntailed()` incorrectly returns `ESat.TRUE` when the absolute-value result is instantiated to `0` and the operand has any domain of size two containing `0`.

This affects reification and negation of `absolute`: Choco can silently omit valid solutions or declare a satisfiable model unsatisfiable. The issue occurs with very small values and does not involve overflow.

The same entailment code is present in both `PropAbsolute` and `PropAbsoluteLight`.

## To reproduce

```java
public class ReifiedAbsoluteZeroBug {
        Model model = new Model();

        IntVar base = model.intVar("base", new int[]{-2, 0});
        IntVar result = model.intVar("result", new int[]{0, 1});
        BoolVar truth = model.absolute(result, base).reify();

        // Make the issue deterministic and independent of the default search strategy.
        model.getSolver().setSearch(
                Search.inputOrderLBSearch(result, base, truth)
        );

        int count = 0;
            count++;
            System.out.printf("base=%d, result=%d, truth=%d%n",
                    base.getValue(), result.getValue(), truth.getValue());
        }
        System.out.println("solution count = " + count);
    }
}
```

Observed output:

```text
base=0, result=0, truth=1
base=-2, result=1, truth=0
base=0, result=1, truth=0
solution count = 3
```

The valid solution below is missing:

```text
base=-2, result=0, truth=0
```

Indeed, `0 != |-2|`, so the reified constraint must be false for this assignment.

The same entailment error can also produce a false UNSAT without relying on a search order:

```java
Model model = new Model();
IntVar result = model.intVar("result", new int[]{0, 1});
IntVar base = model.intVar("base", new int[]{-2, 0});

model.absolute(result, base).getOpposite().post();
model.arithm(result, "=", 0).post();

int count = 0;
while (model.getSolver().solve()) count++;

// Expected: 1 (base = -2)
// Actual:   0
System.out.println(count);
```

## Expected behavior

The first example must return all four assignments:

```text
base=-2, result=0, truth=0
base=-2, result=1, truth=0
base= 0, result=0, truth=1
base= 0, result=1, truth=0
```

The negated example must return one solution with `base=-2`.

## Possible cause

Both propagators contain this entailment condition:

```java
vars[1].getDomainSize() == 2
        && vars[1].contains(vars[0].getValue())
        && vars[1].contains(-vars[0].getValue())
```

It is intended to detect that the operand domain is exactly `{result, -result}`. When `result == 0`, however, `result` and `-result` are the same integer. The condition degenerates into "a domain of size two containing zero", which is true for domains such as `{-2, 0}` or `{0, 5}` even though the constraint is not entailed.

## Possible solution

Only return `TRUE` from this branch when the result is strictly positive:

```java
} else if (vars[0].getValue() > 0
        && vars[1].getDomainSize() == 2
        && vars[1].contains(vars[0].getValue())
        && vars[1].contains(-vars[0].getValue())) {
    return ESat.TRUE;
}
```

The same change is required in:

- `PropAbsolute.isEntailed()`
- `PropAbsoluteLight.isEntailed()`

## Environment

- Choco-solver: reproduced on `4.10.18`, `6.0.1`
- Solver modes: normal mode; the duplicated condition is also present and reproduced in the LCG/light propagator

## Additional context

A directly posted, non-reified `absolute` constraint is not affected by this reproducer. The incorrect result appears when entailment is used, notably through reification or the opposite constraint.

The issue is not limited to enumerated domains. It also occurs with a bounded domain:

```java
Model model = new Model();
IntVar base = model.intVar("base", -1, 0, true);
IntVar result = model.intVar("result", new int[]{0, 1});
model.absolute(result, base).reify();

int count = 0;
while (model.getSolver().solve()) count++;

// Expected: 4
// Actual:   3
System.out.println(count);
```

The operand domain does not need to have size two initially. It is sufficient for search to reduce it to a two-value domain containing zero at some node. For example:

```java
Model model = new Model();
IntVar base = model.intVar("base", -5, 0, true);
IntVar result = model.intVar("result", new int[]{0, 1});
model.absolute(result, base).reify();

int count = 0;
while (model.getSolver().solve()) count++;

// Expected: 12
// Actual:   11
System.out.println(count);
```

cprudhom:
close #1233


ArthurGodet (commented):
I suppose the failing tests are for the same reason as #1234 but to be confirmed (as there is also one failing test in org.chocosolver.solver.LimitsTest.testTime ; EDIT : just a question of time to solve a NQueens problem using only arithm constraints -> not caused by this Pull Request). Other than that, the code seems fine to me

cprudhom:
Updating this branch with the last version of `develop` fixes the test in failure.
So, I think this can be merged into `develop`

Issue #1119, opened 2024-12-09
[BUG] Start and end are always enforced at root


IgnaceBleukx:
Hi,

I'm attempting to model a half-reified Cumulative constraint (see sample code below).
However, while the "non-overlapness" of the Cumulative constraints is correctly reified, the link between start and end variables in the tasks is not reified with it.

E.g., the model  with constraints `[e == 1, bv -> Cumulative([s], [3], [e], [1],[1])]` should be satisfied when the Boolean variable is set to False.
However, as the task is created at the toplevel of the constraint model, it will fail because of the `e == 1` assignment.

Is this considered the expected behaviour of the tasks-constructor? And if so, how should I modify my model to properly reflect a truely half-reified Cumulative constraint?

```java
import org.chocosolver.solver.Model;
import org.chocosolver.solver.variables.IntVar;
import org.chocosolver.solver.variables.BoolVar;
import org.chocosolver.solver.variables.Task;

public class CumulativeModel {
    public static void main(String[] args) {
        // Create a Choco model
        Model model = new Model("Cumulative Scheduling Model");

        // Define variables
        IntVar[] start = new IntVar[3];
        IntVar[] end = new IntVar[3];
        int[] duration = {3, 3, 3};
        IntVar[] demand = new IntVar[3];
        IntVar capacity = model.intVar("cap", 1,1);

        for (int i = 0; i < 3; i++) {
            start[i] = model.intVar("start[" + i + "]", 0, 5);
            end[i] = model.intVar("end[" + i + "]", 0, 5);
            demand[i] = model.intVar("demand[" + i +"]", 1,1);
        }

        // Make the tasks
        Task[] tasks = new Task[3];
        for (int i = 0; i < 3; i++) {
            tasks[i] = new Task(start[i], duration[i], end[i]);
        }

        // Define the Boolean variable
        BoolVar bv = model.boolVar("bv");
        // Add cumulative constraint wrapped in implication
        model.cumulative(tasks, demand, capacity).impliedBy(bv);

        // Add the constraint end[0] == 1
        model.arithm(end[0], "=", 1).post(); // This makes the model UNSAT

        // Solve the model
        if (model.getSolver().solve()) {
            System.out.println("Solution found:");
            for (int i = 0; i < 3; i++) {
                System.out.println("start[" + i + "] = " + start[i].getValue());
                System.out.println("end[" + i + "] = " + end[i].getValue());
            }
        } else {
            System.out.println("No solution found.");
        }
    }
[...]

PR #1237, opened 2026-08-22
Fix #1235

### LINKED ISSUE #1235: [BUG] absolute() posts a constraint as a side effect when the result is instantiated to zero
Related to #1233  , but caused by an independent bug.

## Describe the bug

In Choco 6.x, calling `Model.absolute(result, base)` with `result` instantiated to `0` immediately posts `base = 0` as a side effect, before the returned constraint is posted.

Internal posting is sometimes legitimate in Choco factories when implementing a decomposition through auxiliary variables and channeling constraints. This case is different: the factory posts `base = 0` directly on a user-provided variable. That hard constraint is semantically equivalent to the entire returned `absolute(0, base)` constraint, so it escapes any later reification or negation and makes those operations incorrect.

Merely constructing and reifying the constraint can therefore make an otherwise satisfiable model unsatisfiable.

This appears to be a regression from Choco 4.10.18, where `absolute()` did not contain this preprocessing branch.

## To reproduce

```java
public class AbsoluteFactorySideEffectBug {
        Model model = new Model();

        IntVar result = model.intVar(0);
        IntVar base = model.intVar("base", new int[]{-1, 1});

        // This should only create and reify: result == |base|.
        // Since result is 0 and base is +/-1, the reification Boolean
        // should simply be false for both values of base.
        model.absolute(result, base).reify();

        int count = 0;
            count++;
            System.out.println("base=" + base.getValue());
        }

        System.out.println("solution count = " + count);
    }
}
```

Observed output:

```text
solution count = 0
```

The side effect can also be observed directly, before solving or posting the returned constraint:

```java
Model model = new Model();
model.absolute(model.intVar(0), model.intVar("base", -2, 2));
System.out.println(model.getNbCstrs());
```

Observed output:

```text
1
```

The caller did not post any constraint, but `base = 0` has already been posted internally.

The problem is not specific to `reify()`. Negating the returned constraint also gives a false UNSAT:

```java
Model model = new Model();
IntVar base = model.intVar("base", new int[]{-1, 1});

model.absolute(model.intVar(0), base).getOpposite().post();

int count = 0;
while (model.getSolver().solve()) count++;

// Expected: 2 (base = -1 and base = 1 both satisfy 0 != |base|)
// Actual:   0
System.out.println(count);
```

## Expected behavior

The model contains no hard requirement that `result == |base|`; the constraint is only reified. It must therefore have two solutions:

```text
base=-1  (reification Boolean = false)
base= 1  (reification Boolean = false)
solution count = 2
```

The preprocessing must remain under the control of the returned constraint so that posting, reification and negation preserve their normal semantics. It must not impose the full relation as a hard constraint directly on a user-provided variable before the caller decides how to use the returned constraint.

## Possible cause

`IIntConstraintFactory.absolute()` contains:

```java
} else if (var1.isInstantiated()) {
    if (var1.getValue() == 0) {
        var2.eq(0).post();
    } else if (var1.getValue() > 0) {
        return member(var2, new int[]{-var1.getValue(), var1.getValue()});
    } else {
        return ref().falseConstraint();
    }
}
return new Constraint(ConstraintsName.ABSOLUTE, ...);
```

The zero branch posts `var2 = 0` immediately and then falls through to return an `ABSOLUTE` constraint. Unlike auxiliary channeling constraints used by some factory decompositions, this constraint is posted directly on the caller's `var2` and enforces the full relation unconditionally. A later reification or negation can therefore no longer control it.

## Possible solution

Return the equality constraint instead of posting it:

```java
if (var1.getValue() == 0) {
    return arithm(var2, "=", 0);
}
```

This preserves the preprocessing optimization while allowing the caller to post, reify, or negate the returned constraint normally.

## Environment

- Choco-solver: reproduced on `6.0.1`
- Mode: normal mode

## Additional context

The corresponding `absolute()` implementation in tag `4.10.18` directly returned a `PropAbsolute` constraint and did not contain this zero-result side effect, so this appears to be a 6.x regression.

The standard posted usage remains semantically correct:

```java
Model model = new Model();
IntVar base = model.intVar("base", -1, 1);
model.absolute(model.intVar(0), base).post();
```

It returns the single expected solution `base=0`, although the model contains two posted constraints because the factory has already posted `base=0` before the returned `ABSOLUTE` constraint is posted. The incorrect behavior concerns construction without posting, reification and negation.

This issue is independent from the incorrect zero entailment in `PropAbsolute.isEntailed()`

ArthurGodet:
For #1235, the idea is, as suggested in the issue, to return an arithm constraint instead of posting it.

Otherwise, this Pull Request mainly assure that no constraints are posted within factories anymore (unless LCG is activated).


solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
```diff
@@ -262,17 +261,16 @@ default Constraint absolute(IntVar var1, IntVar var2) {
             }
         } else if (var1.isInstantiated()) {
             if (var1.getValue() == 0) {
-                var2.eq(0).post();
+                return arithm(var2, "=", 0);
```
jgFages, line 265: Agreed!

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
```diff
@@ -351,8 +349,11 @@ default Constraint arithm(IntVar var1, String op1, IntVar var2, String op2, int
                     } else {
                         int[] bounds = VariableUtils.boundsForMultiplication(var1, var2);
                         IntVar var4 = ref().intVar(bounds[0], bounds[1]);
-                        ref().times(var1, var2, var4).post();
-                        return arithm(var4, op2, cste);
```
jgFages, line 355: I do not get what is wrong with the previous version

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
```diff
@@ -723,8 +736,11 @@ default Constraint arithm(IntVar var1, String op1, IntVar var2, String op2, IntV
                     // v1 / v2 OP v3
                     int[] bounds = VariableUtils.boundsForDivision(var1, var2);
                     IntVar var4 = ref().intVar(bounds[0], bounds[1]);
-                    ref().div(var1, var2, var4).post();
```
jgFages, line 726: What was wrong here was that when posting x/y it removes 0 from y?

solver/src/main/java/org/chocosolver/solver/variables/Task.java
```diff
@@ -123,11 +123,13 @@ public Task(IntVar s, IntVar d, IntVar e) {
...
             setActive();
             setPassive();
-        } else {
-            this.getModel().post(new Constraint("Task relation", this));
         }
     }
 
+    public void post() {
+        this.getModel().post(new Constraint("Task relation", this));
+    }
```
jgFages, line 131: I have a mixed feeling about this has a Task is also a variable somehow...

Un unposted Task would make no sens to me

@cprudhom this is worth a check for further implications?

ArthurGodet:
The modifications on Task should also fix #1119 (even if the issue was created at the time we still used TaskMonitor).

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
cprudhom, line 1: In several constraint declarations, you propose moving from a state with two constraints – one of which is reversed – to a state with a single constraint resulting from the merging of the propagators.
Originally, this choice was motivated by improved management of reification: the first constraint is active, whilst only the second (ideally a simple one) is returned to be posted or reified.
I am afraid that the merged versions may lose efficiency:

1. $c_1 \land (c_2\iff b)$
2. $(c_1 \land c_2) \iff b$

ArthurGodet: The motivation of these changes was to remove the confusion that users can experience when some constraints are posted behind their back (which happened when a constraint was posted while only the second one was returned to be either reified or posted).
If the question is performance, maybe we should run a benchmark on XCSP3 and MiniZinc instances to check whether it had impacts (but maybe these instances do not rely that much on reification ?). Otherwise, we might consider having methods that kept the former way, but personally I don't find it satisfying as it leads to confusion for users

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
ArthurGodet: The idea is to avoid posting constraints behind the back of the user (which could lead to faulty behaviour, most especially when reifying constraints)

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
ArthurGodet: No, same reason : avoiding posting constraints behind the back of the user

solver/src/main/java/org/chocosolver/solver/variables/Task.java
ArthurGodet: Yes I agree that it might start being weird for Task being constraints and variables. Maybe we need to revamp Task to only be variables ?

That said, the idea behind this change is about reification (more especially for reifying the cumulative constraint).

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
jgFages: Contrary to the division which removes 0, I do not see which faulty behaviour could come from this. Do you have an example ? 

Somehow var4 is created behind the back of the user. I would tend to prefer having the constraint posted because, whether the constraint is posted or reified, var4 is supposed to represent the result of var1 * var2. 

Also, var4 would be instantiated whenever other variables are instantiated. Whereas it would remain a free variable if the constraint is neither posted nor reified (which would be a wrong thing to do anyway, just saying).

Therefore, even for the division case - where I understand that removing 0 might cause issue (discussable) - I would tend to prefer the current implementation (var4 is defined as var1 / var2).

ArthurGodet: Yeah, for times constraint, as we create the var4 variable for the occasion, I agree it is not necessary to merge the constraints as such. I will change back these cases (but, yes, for the div constraint, posting behind the back of the user might lead to faulty behaviour when using reification)

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
```diff
@@ -2022,12 +2031,13 @@ default Constraint keySort(IntVar[][] vars, IntVar[] PERMvars, IntVar[][] SORTED
         }
         Constraint allDiff = ref().allDifferent(PERMvars);
         allDiff.ignore();
-        return new Constraint(ConstraintsName.KEYSORT,
-                ArrayUtils.append(
-                        allDiff.propagators,
-
-                        new Propagator[]{
-                                new PropKeysorting(vars, SORTEDvars, PERMvars, K)}));
+        return new Constraint(
```
cprudhom, line 2040: We could have used `Constraint.merge(...)` here

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
cprudhom, line 1399: We must not specify the constraint (or introduce any additional variables).

cprudhom, line 1608: We must not specify the constraint (or introduce any additional variables).

cprudhom: The only reason a variable is introduced here is because `PropCountVar` supposes that the `value` variable is enumerated. Looking at the code, I can’t think of any particular reasons for this choice.

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
```diff
@@ -361,8 +360,11 @@ default Constraint arithm(IntVar var1, String op1, IntVar var2, String op2, int
                     } else {
                         int[] bounds = VariableUtils.boundsForDivision(var1, var2);
                         IntVar var4 = ref().intVar(bounds[0], bounds[1]);
-                        ref().div(var1, var2, var4).post();
-                        return arithm(var4, op2, cste);
+                        return Constraint.merge(
+                            ConstraintsName.ARITHM,
+                            ref().div(var1, var2, var4),
+                            arithm(var4, op2, cste)
```
cprudhom, line 372: Depending on the operator, we can avoid using an additional constraint here (and in other cases).
For instance, if op2 equals "GT", we can declare var4 with (cste, bounds[1]).

cprudhom (changes requested):
I think we can take this a step further by avoiding any additional constraints on the user’s part (regarding constraints on integers) and by being more clever about how we declare intermediate variables.
Furthermore, there is a suspicious (and, in my view, unnecessary) declaration of an additional variable in `count`.

solver/src/main/java/org/chocosolver/solver/constraints/IIntConstraintFactory.java
ArthurGodet: Would you prefer to ?

PR #1238, opened 2026-08-28
Fix 1231

### LINKED ISSUE #1231: [BUG] Knapsack removes valid solutions and can return false UNSAT or a wrong optimum
### Description

The `knapsack` constraint can remove valid assignments during propagation.

Depending on the model, this can result in:

- valid solutions being lost;
- a satisfiable model being reported unsatisfiable;
- an incorrect optimum being returned.

The problem is reproducible with small 0/1 instances and small positive integer weights and profits.

---

### Example 1: a valid solution is removed

```java

public class KnapsackLostSolution {

        Model model = new Model();

        IntVar[] occurrences = model.intVarArray("item", 3, 0, 1);
        IntVar weight = model.intVar("weight", 0, 2);
        IntVar profit = model.intVar("profit", 0, 6);

        model.knapsack(
                occurrences,
                weight,
                profit,
                new int[]{1, 1, 3},
                new int[]{1, 2, 3}
        ).post();

        model.arithm(profit, ">=", 2).post();

        while (solver.solve()) {
            System.out.printf(
                    "[%d, %d, %d] weight=%d profit=%d%n",
                    occurrences[0].getValue(),
                    occurrences[1].getValue(),
                    occurrences[2].getValue(),
                    weight.getValue(),
                    profit.getValue()
            );
        }
    }
}
```

Two assignments are feasible:

```text
[0, 1, 0] -> weight=1, profit=2
[1, 1, 0] -> weight=2, profit=3
```

Both satisfy:

```text
weight <= 2
profit >= 2
```

Expected solutions:

```text
[0, 1, 0] weight=1 profit=2
[1, 1, 0] weight=2 profit=3
```

Actual result:

```text
[0, 1, 0] weight=1 profit=2
```

The valid assignment `[1,1,0]` is missing.

#### Invalid deduction during root propagation

The problem can already be observed after a single root propagation:

```java
solver.propagate();

System.out.println(occurrences[0]);
System.out.println(occurrences[1]);
System.out.println(occurrences[2]);
```

Observed domains:

```text
item[0] = 0
item[1] = 1
item[2] = 0
```

However, `item[0]=1` belongs to the valid assignment `[1,1,0]` and must not be removed.

---

### Example 2: wrong optimum with a profit lower bound

Using a fresh model with the same instance as Example 1, set `profit` as the objective:

```java
model.setObjective(Model.MAXIMIZE, profit);

int best = Integer.MIN_VALUE;

while (solver.solve()) {
    best = profit.getValue();
}

System.out.println("Best profit: " + best);
```

The exact optimum is:

```text
[1, 1, 0] -> weight=2, profit=3
```

Expected optimum:

```text
3
```

Actual optimum returned by Choco:

```text
2
```

Therefore, the removed solution can also cause an incorrect optimization result.

---

### Example 3: satisfiable model reported UNSAT

```java

public class KnapsackFalseUnsat {

        Model model = new Model();

        IntVar[] occurrences = model.intVarArray("item", 3, 0, 1);
        IntVar weight = model.intVar("weight", 0, 17);
        IntVar profit = model.intVar("profit", 0, 29);

        model.knapsack(
                occurrences,
                weight,
                profit,
                new int[]{16, 1, 25},
                new int[]{6, 3, 20}
        ).post();

        model.arithm(profit, ">=", 9).post();

        boolean solutionFound = false;

        while (solver.solve()) {
            solutionFound = true;

            System.out.printf(
                    "[%d, %d, %d] weight=%d profit=%d%n",
                    occurrences[0].getValue(),
                    occurrences[1].getValue(),
                    occurrences[2].getValue(),
                    weight.getValue(),
                    profit.getValue()
            );
        }

        if (!solutionFound) {
            System.out.println("No solution");
        }
    }
}
```

The following assignment is valid:

```text
occurrences = [1, 1, 0]

weight = 16*1 + 1*1 + 25*0 = 17
profit =  6*1 + 3*1 + 20*0 = 9
```

It satisfies:

```text
weight <= 17
profit >= 9
```

Expected result:

```text
[1, 1, 0] weight=17 profit=9
```

Actual result:

```text
No solution
```

---

### Additional test: different profit thresholds

For the instance from Example 3, the expected solution counts can be computed by exhaustive enumeration.

There are four assignments satisfying the capacity:

```text
[0,0,0] -> weight=0,  profit=0
[0,1,0] -> weight=1,  profit=3
[1,0,0] -> weight=16, profit=6
[1,1,0] -> weight=17, profit=9
```

The expected and actual solution counts are:

| Minimum profit | Expected | Returned by Choco |
|---:|---:|---:|
| 0 | 4 | 4 |
| 1 | 3 | 3 |
| 2 | 3 | 3 |
| 3 | 3 | 3 |
| 4 | 2 | 2 |
| 5 | 2 | 2 |
| 6 | 2 | 2 |
| 7 | 1 | 0 |
| 8 | 1 | 0 |
| 9 | 1 | 0 |
| 10 | 0 | 0 |

For every threshold from 7 to 9, the valid assignment `[1,1,0]` is removed and the model is reported unsatisfiable.

---

### Additional test: dependency on the occurrence upper bound

Using the instance from Example 3, the result also changes when only the upper bound of the occurrence variables is changed.

#### Occurrences in `0..1`

```java
model.intVarArray("item", 3, 0, 1);
```

Actual result:

```text
No solution
```

Expected solution:

```text
[1,1,0]
```

#### Occurrences in `0..2`

```java
model.intVarArray("item", 3, 0, 2);
```

Actual result:

```text
No solution
```

Expected solution:

```text
[1,1,0]
```

#### Occurrences in `0..3`

```java
model.intVarArray("item", 3, 0, 3);
```

Choco finds:

```text
[0,3,0]
[1,1,0]
```

The assignment `[1,1,0]` is valid for all three occurrence domains.

---

### Example 4: wrong optimum without an external profit bound

An incorrect optimum can also be reproduced without explicitly posting a minimum-profit constraint.

```java

public class KnapsackWrongOptimum {

        Model model = new Model();

        IntVar[] occurrences = model.intVarArray("item", 5, 0, 1);
        IntVar weight = model.intVar("weight", 0, 14);
        IntVar profit = model.intVar("profit", 0, 25);

        model.knapsack(
                occurrences,
                weight,
                profit,
                new int[]{1, 6, 7, 9, 5},
                new int[]{1, 6, 8, 9, 1}
        ).post();

        model.setObjective(Model.MAXIMIZE, profit);
        model.getSolver().setSearch(
                Search.randomSearch(occurrences, 3)
        );

        int best = Integer.MIN_VALUE;

            best = profit.getValue();
        }

        System.out.println("Best profit: " + best);
    }
}
```

The exact optimum is 15:

```text
occurrences = [1,1,1,0,0]
weight = 1 + 6 + 7 = 14
profit = 1 + 6 + 8 = 15
```

Expected optimum:

```text
15
```

Actual optimum returned by Choco with random-search seed `3`:

```text
14
```

#### Additional random-search seeds

The same instance was tested with every random-search seed from `0` to `19`.

The incorrect optimum `14` is returned with seeds:

```text
3, 8, 9
```

The correct optimum `15` is returned with the other tested seeds:

```text
0, 1, 2, 4, 5, 6, 7, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19
```

No time limit or other stopping criterion was used. The search strategy changes the order in which solutions are explored, but it should not change the final optimum.

---

### Additional diagnostic: behavior with LCG enabled

As an additional diagnostic, the examples were also tested by replacing:

```java
Model model = new Model();
```

with:

```java

Model model = new Model(
        "My problem",
        SettingsBuilder.init().setLCG(true).build()
);
```

With this setting:

- Example 1 returns both valid solutions;
- Example 2 returns the correct optimum `3`;
- Example 3 finds the valid assignment `[1,1,0]`;
- Example 4 returns the correct optimum `15` with random-search seed `3`, `8`, `9`.

The problem therefore does not reproduce with LCG enabled.

Looking at `IIntConstraintFactory#knapsack`, `PropKnapsackKatriel01` is included in the default configuration but is not included in the LCG configuration.

As another diagnostic check, removing only `PropKnapsackKatriel01` locally from the default configuration also makes the examples return the expected results.

These observations suggest that the incorrect filtering is related to `PropKnapsackKatriel01`. I have not investigated its internal algorithm further.

---

### Environment

- Choco Solver: `6.0.1`
- Java: OpenJDK 26.0.1

cprudhom:
Branch fix_1231 brings the following improvements to the Knapsack module:

   1. Bug fix (close #1231): Synchronize search trees before first propagation in PropKnapsackKatriel01
   2. Complete documentation: Added Javadoc for:
      • Propagators (PropKnapsack, PropKnapsackKatriel01)
      • Data structures (13 files)
   3. Test reinforcement: +492 lines of tests for KnapsackTest.java
   4. Readability improvement: Renamed variables to match academic terminology

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/constraints/nary/knapsack/PropKnapsackKatriel01.java
+++ b/solver/src/main/java/org/chocosolver/solver/constraints/nary/knapsack/PropKnapsackKatriel01.java
@@ -25,41 +25,175 @@ import java.util.ArrayList;
 import java.util.Arrays;
 
 /**
- * Propagator for the 0/1-Knapsack constraint
- * based on Dantzig-Wolfe relaxation trying
- * to find forbidden and mandatory items
+ * Propagator for the 0/1-Knapsack constraint using cost-based filtering.
+ * <p>
+ * This propagator implements the algorithm described in:
+ * <ul>
+ *   <li>Fahle, T., & Sellmann, M. (2002). Cost Based Filtering for the Constrained Knapsack Problem.
+ *       Annals of Operations Research, 115, 73-93.</li>
+ *   <li>Katriel, I., Sellmann, M., Upfal, E., & Van Hentenryck, P. (2007).
+ *       Propagating Knapsack Constraints in Sublinear Time. AAAI 2007, 231-236.</li>
+ * </ul>
+ * <p>
+ * The algorithm performs cost-based filtering by:
+ * <ol>
+ *   <li>Sorting items by decreasing efficiency (profit/weight ratio), breaking ties in favor of larger weights.</li>
+ *   <li>Computing the Dantzig relaxation to find the critical item (first item that would exceed capacity).</li>
+ *   <li>Using two specialized search trees:
+ *       <ul>
+ *         <li>{@code ItemFindingSearchTree}: To efficiently find the next item to check, based on weight.</li>
+ *         <li>{@code ComputingLossWeightTree}: To compute the maximum weight that can be replaced
+ *             without violating the profit bound.</li>
+ *       </ul>
+ *   </li>
+ *   <li>Identifying mandatory items (items that must be included in any solution) and
+ *       forbidden items (items that cannot be included in any solution).</li>
+ * </ol>
+ * <p>
+ * The propagator maintains the following invariant:
+ * For a given lower bound B on the objective (profit), an item x_i is:
+ * <ul>
+ *   <li><b>Mandatory</b> if the fractional optimum of (X \ {x_i}, C) &lt; B.</li>
+ *   <li><b>Forbidden</b> if the fractional optimum of (X \ {x_i}, C - w_i) + p_i &lt; B.</li>
+ * </ul>
+ * <p>
+ * This implementation uses an incremental approach with amortized linear time complexity
+ * per call after an O(n log n) preprocessing step.
  *
  * @author Nicolas PIERRE
+ * @author Charles Prud'homme
  */
 public class PropKnapsackKatriel01 extends Propagator<IntVar> {
+    
+    /**
+     * State constant indicating that an item has been added to the solution.
+     * An item in this state is included in every valid solution and its value 0
+     * has been removed from the variable domain.
+     */
     static final int ADDED = 1;
+    
+    /**
+     * State constant indicating that an item has been removed from the problem.
+     * An item in this state cannot be included in any valid solution and its value 1
+     * has been removed from the variable domain.
+     */
     static final int REMOVED = -1;
+    
+    /**
+     * State constant indicating that an item's status is not yet determined.
+     * Items in this state may or may not be included in the optimal solution.
+     */
     static final int NOT_DEFINED = 0;
     // ***********************************************************************************
     // VARIABLES
     // ***********************************************************************************
 
+    /**
+     * Array mapping original item indices to their position in the efficiency-sorted order.
+     * Items are sorted by decreasing efficiency (profit/weight), with ties broken in favor of larger weights.
+     */
     private final int[] order;
+    
+    /**
+     * Array mapping efficiency-sorted indices back to original item indices.
+     * This is the inverse of {@link #order}.
+     */
     private final int[] reverseOrder;
+    
+    /**
+     * Number of items in the knapsack problem.
+     */
     private final int n;
+    
+    /**
+     * The capacity variable of the knapsack (maximum allowed weight).
+     * The upper bound of this variable represents the available capacity.
+     */
     private final IntVar capacity;
+    
+    /**
+     * The power (profit) variable of the knapsack (total profit to maximize).
+     * The lower bound of this variable represents the minimum required profit.
+     */
     private final IntVar power;
+    
+    /**
+     * State of each item: {@link #NOT_DEFINED}, {@link #ADDED}, or {@link #REMOVED}.
+     * This array tracks which items have been forced into or out of the solution.
+     */
     private final int[] itemState;
+    
+    /**
+     * Search tree for efficiently finding the next item to check based on weight.
+     * Used to implement the monotonicity property: when processing items in weight order,
+     * the critical item position increases monotonically.
+     */
     private final ItemFindingSearchTree findingTree;
+    
+    /**
+     * Search tree for computing the maximum weight that can be replaced without violating
+     * the profit bound. Contains all items sorted by efficiency and stores cumulative weight
+     * and profit sums in internal nodes.
+     */
     private final ComputingLossWeightTree computingTree;
+    
+    /**
+     * Information about the critical item from the Dantzig relaxation:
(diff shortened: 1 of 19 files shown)


cprudhom:
#1237 must be amended accordingly

solver/src/main/java/org/chocosolver/solver/constraints/nary/knapsack/PropKnapsackKatriel01.java
jgFages: full and not custom ?

solver/src/main/java/org/chocosolver/solver/constraints/nary/knapsack/PropKnapsackKatriel01.java
```diff
@@ -111,29 +253,47 @@ public int getPropagationConditions(int vIdx) {
...
-            // updates on the energy variable
+            // updates on the total profit variable
             return IntEventType.lowerBoundAndInst();
         }
     }
 
     @Override
     public void propagate(int evtmask) throws ContradictionException {
+        if (PropagatorEventType.isFullPropagation(evtmask)) {
+        // Synchronize trees with current variable states BEFORE computing critical item.
```
jgFages, line 264: computingTree and findingTree do not need to be cleared before continuing ?

solver/src/main/java/org/chocosolver/solver/constraints/nary/knapsack/PropKnapsackKatriel01.java
cprudhom: You are right. This is legacy but not correct.

solver/src/main/java/org/chocosolver/solver/constraints/nary/knapsack/PropKnapsackKatriel01.java
cprudhom: This is achieved incrementally thanks to methods like `removeItemFromProblem`, `addItemToSolution` and their opposites.

ArthurGodet (changes requested):
For me, modifications are OK, but as @jgFages said, I agree there is a forcePropagate() to change.

That said, the failing test should also be investigated on why it is failing

ArthurGodet:
> #1237 must be amended accordingly

Done 😉

cprudhom:
I’ve just spent a bit of time looking into what the problem is with the test that’s failing.
At first glance, it seems that the patch provided here is necessary but not sufficient. I’ll try to put together a minimal working example, but it seems that despite a (necessary) update to the structures, there’s a deeper issue.

Given the complexity of the algorithm and the underlying structures, the fact that the original contributor cannot be easily contacted, and the time I have already spent on this issue, I am reluctant to look into it further. 

I propose amending this pull request to add a comment explaining the use of the constraint in question and opening a separate issue to describe the bug.

cprudhom:
I plunged back into the 2nd error and it seems that there is a bug in the neighbourhood. 
I'll double-check in the next days.

PR #1243, opened 2026-09-03
Fix issue #1242 : handles the case of a bounded variable differently

### LINKED ISSUE #1242: [BUG] Unsound reified set membership on bounded domains
## Describe the bug

In Choco 6.0.1, the public `IReificationFactory` methods `reifyXinS(X, S, B)`
and `reifyXnotinS(X, S, B)` may accept assignments that violate their
documented equivalences when `X` has a bounded domain and the set that must be
enforced is non-contiguous.

For example, `reifyXinS(x, {0,2}, true)` may accept `x=1`. The symmetric
`reifyXnotinS(x, {1}, true)` is affected as well: it may also accept `x=1`.

The issue reproduces with the small domain `[0,2]`. It does not require LCG or
large integer values.

## To reproduce

```java

public class ReifiedSetMembershipMWE {
        Model model = new Model();

        IntVar x = model.intVar("x", 0, 2, true); // bounded domain
        BoolVar b = model.boolVar("b");
        IntIterableRangeSet set =
                new IntIterableRangeSet(new int[]{0, 2});

        model.reifyXinS(x, set, b);
        model.arithm(b, "=", 1).post();
        model.getSolver().setSearch(Search.inputOrderLBSearch(x));

            System.out.printf("x=%d, b=%d%n",
                    x.getValue(), b.getValue());
        }
    }
}
```

The first solution is valid:

```text
x=0, b=1
```

The next search branch reaches `x=1`. With the default solution checker,
execution then ends with an `InvalidSolutionException` identifying the
`BASIC_REIF` constraint and its `PropXinSHalfReif` propagators. The reported
search path is equivalent to:

```text
ROOT, x!=0, x=1
```

## `reifyXnotinS` variant

The same MWE reproduces through the other public method by replacing the set
construction and the `reifyXinS` call with:

```java
IntIterableRangeSet set = new IntIterableRangeSet(new int[]{1});

model.reifyXnotinS(x, set, b);
```

Here, `b=1` means `x ∉ {1}`, so only `x=0` and `x=2` are valid.
Nevertheless, the solver can reach `x=1, b=1` in the same way.

## Expected behavior

For both versions above, the only solutions are:

```text
x=0
x=2
```

The bounded domain is not expected to represent the hole at `1` during root
propagation. Keeping the interval `[0,2]` is valid weaker filtering, but the
constraint must remain able to reject `x=1` when later bound decisions
instantiate the variable to that value.

## Comparison with general reification

The following equivalent formulation does not reproduce the issue:

```java
model.member(x, new int[]{0, 2}).reifyWith(b);
model.arithm(b, "=", 1).post();
```

It returns only `x=0` and `x=2` with the same bounded variable and search. Thus,
the documented semantics can be enforced correctly on a bounded domain; the
failure is specific to the specialized `reifyXinS` / `reifyXnotinS` path.

Changing `x` to an enumerated domain also does not reproduce this MWE.

## FlatZinc entry point

The specialized path is also used internally by the FlatZinc parser. For a
literal set, `set_in_reif` creates an `IntIterableRangeSet` and calls
`model.reifyXinS(...)`.

For example, with the default automatic domain representation threshold:

```fzn
var 0..65535: x :: output_var;

constraint set_in_reif(x, {0, 65535}, true);

solve :: int_search([x], input_order, indomain_min, complete) satisfy;

output ["x=", show(x), "\n"];
```

Run it with all-solution search enabled:

```text
java -cp choco-solver-6.0.1-light.jar org.chocosolver.parser.flatzinc.ChocoFZN -a model.fzn
```

Choco prints the valid solution `x=0`, then reports an
`InvalidSolutionException` on the branch `x!=0, x=1`. This reaches the affected
propagator without user code directly constructing an `IntIterableRangeSet`.
Without `-a`, the FlatZinc parser stops after the first solution and the invalid
second branch is not observed.

## Observed propagator behavior

When its Boolean is true, `PropXinSHalfReif` iterates over the values outside
its allowed set and calls `removeValue(...)`, then becomes passive. For a
bounded domain, removal of an interior value such as `1` from `[0,2]` cannot be
represented and leaves the interval unchanged. The propagator nevertheless
becomes passive, allowing that value to be selected later by bound decisions.

In 6.0.1, both public methods install `PropXinSHalfReif` instances, with the
Boolean and the complementary set arranged according to the requested
equivalence, which is why both entry points may be affected.

## Environment

- Choco-solver: `6.0.1`
- Solving mode: normal mode, LCG disabled
- JRE: OpenJDK Zulu `26.0.1`
- OS: Microsoft Windows `10.0.19045`

cprudhom:
+ update javadoc

close #1242

### DIFF
--- a/solver/src/main/java/org/chocosolver/solver/constraints/reification/PropXinSHalfReif.java
+++ b/solver/src/main/java/org/chocosolver/solver/constraints/reification/PropXinSHalfReif.java
@@ -20,16 +20,35 @@ import org.chocosolver.util.objects.setDataStructures.iterable.IntIterableRangeS
 /**
  * A propagator dedicated to express in a compact way: b &rArr; (x &isin; c)
  * <br/>
+ * This propagator ensures that if the boolean variable b is true, then the integer variable x must belong to the set c.
  *
  * @author Charles Prud'homme
  * @since 12/03/2024
  */
 @Explained
 public class PropXinSHalfReif extends Propagator<IntVar> {
+    /**
+     * The integer variable that must belong to the set when b is true.
+     */
     final IntVar var;
+    
+    /**
+     * The set of allowed values for the integer variable when b is true.
+     */
     final IntIterableRangeSet set;
+    
+    /**
+     * The boolean variable that controls the reification: when true, x must be in the set.
+     */
     final BoolVar b;
 
+    /**
+     * Creates a reified propagator for x in set.
+     *
+     * @param x    the integer variable to constrain
+     * @param set  the set of allowed values for x when b is true
+     * @param b    the boolean variable that controls the reification
+     */
     public PropXinSHalfReif(IntVar x, IntIterableRangeSet set, BoolVar b) {
         super(new IntVar[]{x, b}, PropagatorPriority.BINARY, false, true);
         this.set = set.duplicate();
@@ -40,13 +59,19 @@ public class PropXinSHalfReif extends Propagator<IntVar> {
     @Override
     public int getPropagationConditions(int vIdx) {
         if (vIdx == 0) {
+            // var (IntVar) requires all event types for propagation
             return IntEventType.all();
         }
+        // b (BoolVar) only requires instantiation events
         return IntEventType.INSTANTIATE.getMask();
     }
 
     @Override
     public void propagate(int evtmask) throws ContradictionException {
+        // Propagation logic for b -> (x in set) constraint
+        // When b is false, no filtering is required on x
+        // When b is true, x must be filtered to only values in the set
+        // When x is out of set bounds, b must be false
         if (b.isInstantiatedTo(0)) { // if b is false, then no filtering is required
             setPassive();
             return;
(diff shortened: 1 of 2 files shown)


solver/src/main/java/org/chocosolver/solver/constraints/reification/PropXinSHalfReif.java
```diff
@@ -55,30 +80,60 @@ public void propagate(int evtmask) throws ContradictionException {
...
+                    if (!set.contains(i)) {
+                        var.removeValue(i, this, lcg() ?
+                                this.r(b.getValLit()) : Reason.undef());
+                    }
+                }
+                setPassive();
+            } else {
+                int lb = var.getLB();
+                int i = 0;
+                while (!set.contains(lb + i)) {
```
jgFages: && lb+i <= ub 
For performance in case the set is large?
Idem for the other while loop

ArthurGodet (approved):

PR #1245, opened 2026-09-06
Fix integer overflow when composing nested affine views

### LINKED ISSUE #1244: [BUG] Integer overflow when composing nested affine views produces incorrect results
## Describe the bug

Composing two affine integer expressions can overflow while Choco flattens the
nested views, even when every value that the expression can actually take is
well within Choco's recommended integer bounds.

The corrupted view can affect correctness in both directions:

- a valid solution can be removed and the model reported unsatisfiable;
- an assignment that violates the posted constraint can be returned as a
  solution.

The issue is reproducible through the public arithmetic-expression API, with
views enabled (the default).

## To reproduce

The following complete example demonstrates both effects:

```java
import org.chocosolver.solver.Model;
import org.chocosolver.solver.variables.IntVar;

public class NestedAffineViewOverflow {

        falseUnsat();
        invalidSolution();
    }

    private static void falseUnsat() {
        Model model = new Model();
        IntVar x = model.intVar("x", 46_341, 46_342, false);
        IntVar result = x.sub(46_341).mul(46_341).intVar();

        result.eq(0).post();

        boolean solved = model.getSolver().solve();
        System.out.println("eq(0): " + solved);
        // Actual:   false
        // Expected: true, with x = 46_341
    }

    private static void invalidSolution() {
        Model model = new Model();
        IntVar x = model.intVar("x", 46_341, 46_342, false);
        IntVar result = x.sub(46_341).mul(46_341).intVar();

        result.ne(0).post();

            System.out.println("ne(0): x = " + x.getValue() + ", result = " + result.getValue());
        }
        // Actual:   x = 46341, result = 0; then x = 46342, result = 46341
        // Expected: x = 46342 only
    }
}
```

The first example uses the default `Model` settings. It prints:

```text
eq(0): false
```

The second example also uses the default `Model` settings and prints:

```text
ne(0): x = 46341, result = 0
ne(0): x = 46342, result = 46341
```

Changing the first model to use

```java
new Model(SettingsBuilder.init().setEnableViews(false))
```

makes it find the expected solution `x = 46_341`.

The problem occurs with both bounded and enumerated domains for `x`.

## Expected behavior

The expression has only two possible mathematical values:

```text
x = 46341 -> (46341 - 46341) * 46341 = 0
x = 46342 -> (46342 - 46341) * 46341 = 46341
```

Therefore:

- `result.eq(0)` must have the single solution `x = 46_341`;
- `result.ne(0)` must have the single solution `x = 46_342`.

Both possible expression values, `0` and `46_341`, are well below
`IntVar.MAX_INT_BOUND` (`21_474_836`). The model itself does not require an
out-of-range value.

## Scope

Targeted checks show that the same corrupted affine view affects all six
relational operators: `=`, `!=`, `<`, `<=`, `>` and `>=`. Depending on the
operator, a valid value can be removed, an invalid value can be kept, or the
default solution checker can raise `InvalidSolutionException` after propagation
accepts an invalid assignment.

The `!=` case in the main reproducer is particularly important: with the
default settings, Choco returns `x = 46_341` even though `result = 0`, without
raising an exception or emitting a warning. The solution checker evaluates the
constraint through the same corrupted view and does not detect the violation.

For example, the following model has the valid solution `x = 46_342`, for which
`result = 46_341`:

```java
Model model = new Model();
IntVar x = model.intVar("x", 46_341, 46_342, false);
IntVar result = x.sub(46_341).mul(46_341).intVar();
model.arithm(result, ">=", 1).post();

model.getSolver().solve();
```

Instead of returning that solution, `solve()` raises
`InvalidSolutionException`. The corrupted view first leads the solver to the
invalid assignment `x = 46_341`, which is then rejected by the default solution
checker.

With the solution checker disabled by the production settings, the same
propagation error becomes directly observable in the returned solutions:

```java
Model model = new Model(SettingsBuilder.prod());
IntVar x = model.intVar("x", 46_341, 46_342, false);
IntVar result = x.sub(46_341).mul(46_341).intVar();
model.arithm(result, ">=", 1).post();

    System.out.println("x = " + x.getValue() + ", result = " + result.getValue());
}
```

The output is:

```text
x = 46341, result = 0
x = 46342, result = 46341
```

Only `x = 46_342` is valid. For `x = 46_341`, the expression is equal to zero
and therefore violates `result >= 1`.

The observed results for the same expression are:

| Constraint | Expected `x` | Default settings | `prod()` settings |
|---|---:|---:|---:|
| `result = 0` | `[46341]` | `[]` | `[]` |
| `result != 0` | `[46342]` | `[46341, 46342]` | `[46341, 46342]` |
| `result < 1` | `[46341]` | `[]` | `[]` |
| `result <= 0` | `[46341]` | `[]` | `[]` |
| `result > 0` | `[46342]` | `InvalidSolutionException` | `[46341, 46342]` |
| `result >= 1` | `[46342]` | `InvalidSolutionException` | `[46341, 46342]` |

## Possible cause

The nested affine view is flattened in `IViewFactory.intView(...)`. The
composed coefficients are currently calculated as `int`:

```java
int av = (view.p ? 1 : -1) * view.a * a;
int bv = a * view.b + b;
```

For `(x - 46_341) * 46_341`, the expanded constant is:

```text
-46341 * 46341 = -2,147,488,281
```

This is only `4,633` below `Integer.MIN_VALUE`, so the `int` calculation wraps
to:

```text
2,147,479,015
```

The overflow concerns the constant of the expanded representation, not a value
that the original expression can take on its domain.

`46_341` is also a compact boundary case: `46_340 * 46_340 = 2,147,395,600`
still fits in an `int`, whereas `46_341 * 46_341 = 2,147,488,281` does not.

## Possible solution

Calculating only the composed coefficients in `long` and checking `av` and
`bv` is not sufficient. For example:

```java
IntVar x = model.intVar("x", 46_273, 46_276, false);
IntVar result = x.sub(46_273).mul(46_406).intVar();
result.eq(139_218).post();
```

The valid solution is `x = 46_276`, but Choco returns no solution. The flattened
coefficients are both representable:

```text
av = 46,406
bv = -2,147,344,838
```

However, `av * 46,276 = 2,147,484,056` is not representable as an `int`.
During inverse propagation, the equivalent calculation
`139,218 - (-2,147,344,838)` reaches the same value and overflows before the
division by `av`.

One possible approach is therefore to check both the composed coefficients and
the scaled and shifted endpoint values before flattening. Otherwise, the
original view can be kept nested:

```java
static boolean fitsInt(long value) {
    return value >= Integer.MIN_VALUE && value <= Integer.MAX_VALUE;
}

long lav = (view.p ? 1L : -1L) * view.a * a;
long lbv = (long) a * view.b + b;
if (!fitsInt(lav) || !fitsInt(lbv)) {
    return new IntAffineView<>(var, a, b);
}

IntVar base = view.getVariable();
long scaledLB = lav * base.getLB();
long scaledUB = lav * base.getUB();
if (!fitsInt(scaledLB) || !fitsInt(scaledUB)
        || !fitsInt(scaledLB + lbv) || !fitsInt(scaledUB + lbv)) {
    return new IntAffineView<>(var, a, b);
}

int av = (int) lav;
int bv = (int) lbv;
```

This fallback uses the same `new IntAffineView<>(var, a, b)` construction that
is already used by the current non-flattening branch; it does not require a new
view type.

No remaining correctness failure was observed in the tested scope. The extended
guard fixes all identified counterexamples and the targeted checks for all six
relational operators. The existing Choco view tests also pass with this guard.
This remains a suggested implementation; a more efficient solution may be
possible while preserving the same view invariants.

## Environment

- Choco-solver: `6.0.1`
- JRE used for the reproduction: OpenJDK `26.0.1`
- LCG: not enabled

adityaanikam:
Fixes #1244.

`IViewFactory#intView` flattens a nested affine view by composing the coefficients in `int` arithmetic:

```java
int av = (view.p ? 1 : -1) * view.a * a;
int bv = a * view.b + b;
```

Either product can leave the `int` range even when every value the expression can take is small. In the reported case, `x` ranges over `[46341, 46342]` and the composed expression is at most `46341`, but flattening `(x - 46341) * 46341` computes `bv` as `46341 * -46341 = -2147488281`, which underflows and wraps to `+2147479015`. The view built from that is wrong in both directions: a valid solution is pruned and the model reported unsatisfiable, and an assignment that violates the posted constraint is returned as a solution.

Changes proposed in this PR:
- Compose the coefficients in `long` and keep the nested view when either does not fit in an `int`, falling back to the `new IntAffineView<>(var, a, b)` construction the non-flattening branch already used. No new view type is needed and views stay enabled.
- Add two regression tests to `IntAffineViewTest`, one per direction reported in the issue.
- Add myself to the authors of both modified classes, as CONTRIBUTING asks.

On scope: the issue also suggests guarding the scaled bounds of the base variable, in addition to the coefficients. I left that out. It only changes behaviour when `av` and `bv` both fit but `av * base` does not, and in that situation this patch and the current code take the same branch, so I could not write a test that distinguishes them. I would rather not add a condition with no regression test behind it. Worth noting if you do want it covered: `getLB()` and `getUB()` compute `(...) * a + b` as a single `int` expression, so an overflowing intermediate cancels there and the bound is still correct; `contains()` is the one that would suffer, because it does `value -= b` and then divides, and the division does not undo the wrap. Happy to add it if you would like it handled.

I did not touch `CHANGES.md`, since there is no unreleased section to add to.

Verified with a negative control: reverting only `IViewFactory.java` and keeping the tests fails exactly the two new tests out of 150 in the class, with `expected [true] but found [false]` for the pruned solution and `expected [46342] but found [46341]` for the invalid one. Restoring the fix passes all 150.

@chocoteam/core-developer

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,28 +9,10 @@ This file is dedicated to sum up the new features added and bugs fixed in Choco-
 
 ### Major features:
 
-#### Constraints & LCG
-- Improved `element` constraint: new propagator for bounded result variables and optimized handling of large arrays with repeated values
-- Improved `modulo` constraint: faster propagation for both binary and ternary cases; fixed handling of negative values in expressions and `mod` with negatives
-- Improved `div` constraint (`PropDivXYZ`): better filtering based on sign analysis of variables
-- LCG: added `extractFromVariables` setting to extract nogoods from variable domains (rather than decision path) on solution (#1193)
-- Fixed `PropModXY` incorrect filtering
-- Fixed `PropElementIn` propagator
-- Fixed `SparseBitSet.prevClearBit()` and unified its implementation with `nextClearBit()`
-
-#### Build, CI & Tooling
-- Migrated build and GitHub Actions CI to Java 17
-- Upgraded ANTLR from 4.9.3 to 4.13.2, TestNG from 7.5.1 to 7.12.0, and args4j from 2.33 to 2.37
-- Bumped jgrapht, sizeof, slf4j, and Maven plugin versions
-- Moved `slf4j-nop` to test scope to avoid classpath conflicts in applications depending on choco-solver
-- Centralized dependency versions in POM properties and cleaned up redundant POM configuration
-- Updated release scripts and changelog generation tooling
-
-### Other closed issues and pull requests:
-- #1193: Add a parameter to extract nogoods from variables on solution (LCG)
+#### Constraints
+- Fix PropModXY in case of negative mod
 
 #### Contributors to this release:
-- Arthur Godet <arth.godet@gmail.com>
 - Charles Prud'homme <charles.prudhomme@imt-atlantique.fr>
 - Jean-Guillaume Fages <jg.fages@cosling.com>
 


solver/src/main/java/org/chocosolver/solver/variables/IViewFactory.java
```diff
@@ -128,16 +129,21 @@ default IntVar intView(int a, IntVar var, int b) {
...
-                int bv = a * view.b + b;
-                if (av == 1 && bv == 0) {
-                    return view.getVariable();
-                } else {
-                    return intView(av, view.getVariable(), bv);
+                // Compose in long: flattening can overflow int even when every
+                // value the composed expression can take is well within range.
+                long av = (view.p ? 1L : -1L) * view.a * a;
+                long bv = (long) a * view.b + b;
+                if (fitsInt(av) && fitsInt(bv)) {
```
ArthurGodet: I would move this if condition to the else inside it.

ArthurGodet (changes requested):
Just a small change to do (to keep simplified imbricated if/else-if), but otherwise, changes seem fine to me. Thanks for your contribution !

solver/src/main/java/org/chocosolver/solver/variables/IViewFactory.java
adityaanikam: Done, thanks. You are right that the fitsInt check was redundant in that branch, since av == 1 and bv == 0 always fit.

ArthurGodet (approved):

cprudhom (approved):

